PR #19, opened 2018-05-22
Add block confirmation before transaction validation

### LINKED ISSUE #5: Transaction validation after specific number of confirmed block

patitonar:
Closes #5

### DIFF
--- a/src/processCollectedSignatures.js
+++ b/src/processCollectedSignatures.js
@@ -7,7 +7,7 @@ const {sendRawTx, sendTx} = require('./tx/sendTx');
 const {createMessage, signatureToVRS} = require('./message');
 const {getNonce} = require('./tx/web3');
 const {getGasPrices} = require('./gasPrice');
-const {asyncForEach} = require('./utils');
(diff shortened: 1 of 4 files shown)


src/processCollectedSignatures.js
```diff
@@ -68,6 +73,10 @@ async function processCollSignatures(signatures, foreignChainId){
     let nonce = await getNonce(web3Foreign, VALIDATOR_ADDRESS);
     nonce = Math.max(dbNonce.foreign, nonce);
     await asyncForEach(signatures, async (colSignature, indexSig) => {
+      if(requiredBlockConfirmations > 1) {
```
fvictorio: What is the reason for this check? Even if `requiredBlockConfirmations` is 1, it's useful to call the function because that wait it will wait until the transaction is mined.

patitonar: The event belongs to a transaction that was already mined so I added that check to avoid calling `waitForBlockConfirmations` if only 1 block was needed

PR #21, opened 2018-05-23
Add block confirmation before transaction validation

### LINKED ISSUE #5: Transaction validation after specific number of confirmed block

fvictorio:
Closes #5.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -8,7 +8,12 @@ function getRequiredBlockConfirmations(contract) {
   return contract.methods.requiredBlockConfirmations().call()
 }
 
-function waitForBlockConfirmations(web3, event, requiredBlockConfirmations, blockNumberProvider) {
+function waitForBlockConfirmations({
+  web3,
+  event,
+  requiredBlockConfirmations,
+  blockNumberProvider
+}) {
   const transactionReceiptAsync = async function(event, resolve, reject) {
     const latestBlockNumber = blockNumberProvider.getLatestBlockNumber()
     const blockConfirmations = latestBlockNumber - event.blockNumber + 1
@@ -16,7 +21,18 @@ function waitForBlockConfirmations(web3, event, requiredBlockConfirmations, bloc
       try {
         const receipt = await web3.eth.getTransactionReceipt(event.transactionHash)
         if (receipt && receipt.blockNumber) {
-          resolve(receipt)
+          if (event.blockNumber === receipt.blockNumber) {
+            resolve(receipt)
+          } else {
+            const updatedEvent = {
+              ...event,
+              blockNumber: receipt.blockNumber
+            }
+
+            setTimeout(() => {
+              transactionReceiptAsync(updatedEvent, resolve, reject)
+            }, 5000)
+          }
         } else {
           reject('Tx not found')
         }
--- a/src/utils.js
+++ b/src/utils.js
@@ -8,31 +8,30 @@ function getRequiredBlockConfirmations(contract) {
   return contract.methods.requiredBlockConfirmations().call()
 }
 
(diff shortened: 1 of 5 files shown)


fvictorio:
@patitonar was the one that implemented this, I just reverted the PR and re-created it. I'm still becoming familiar with the code.

How do you do those kind of tests? Is it enough with setting `NUMBER_OF_WITHDRAWALS_TO_SEND` to 100 and running the script?

patitonar:
@fvictorio @rstormsf @akolotov Added some changes to reduce the number of http calls when waiting for block confirmation. How are you testing performance?

rstormsf:
@fvictorio yes, just increment it to 100 and it should send 100 tx. I recommend doing against some localhost nodes

src/utils.js
```diff
@@ -4,6 +4,44 @@ async function asyncForEach(array, callback) {
   }
 }
 
+function getRequiredBlockConfirmations(contract) {
+  return contract.methods.requiredBlockConfirmations().call()
+}
+
+function waitForBlockConfirmations(web3, event, requiredBlockConfirmations, blockNumberProvider) {
```
rstormsf: the reason why I always use  an object instead of params, so nobody  has to follow the order of params.
Please make it as an object.
```js
{web3, event, requiredBlockConfirmations, blockNumberProvider}
```

src/utils.js
```diff
@@ -4,6 +4,44 @@ async function asyncForEach(array, callback) {
...
+  const transactionReceiptAsync = async function(event, resolve, reject) {
+    const latestBlockNumber = blockNumberProvider.getLatestBlockNumber()
+    const blockConfirmations = latestBlockNumber - event.blockNumber + 1
+    if (blockConfirmations >= requiredBlockConfirmations) {
+      try {
+        const receipt = await web3.eth.getTransactionReceipt(event.transactionHash)
+        if (receipt && receipt.blockNumber) {
```
rstormsf: we should probably also check for `event.blockNumber` === `receipt.blockNumber`
to make sure that original tx still has the same blockNumber as we received it. if not, that means it was already forked and was included in different block so we have to recount our block confirmations

akolotov:
@rstormsf thanks!
This is exactly the test needs to be done to test bridge performance.
Expected behavior for this test: If you have all 100 transaction generating `Deposit()` event in one block, the corresponding `submitSignature` calls should also be included in one or two blocks. Similar is for `CollectedSignatures()` and `Withdraw()` events.

patitonar:
@rstormsf Added changes from review

patitonar:
@rstormsf @akolotov About this PR, we were thinking of another solution. 

Today the process listen to events until the lastest block, if we need 5 blocks confirmation, it waits by checking the last block number until the number of confirmation is reached and call `getTransactionReceipt` to make sure the Tx is there.

Instead of doing that, we can listen to events on blocks until `lastBlock - confirmationNumber`. 
For example, if `5` blocks needed for confirmation, and last block number is `100` we will listen for events until block `96`. In that way, we will only receive confirmed events, and we don't need to do any other check, avoiding the extra work of checking and calling `getTransactionReceipt`.

What do you think of this solution?

patitonar:
We are using the stored block to know from which block listen the next time, so I think in this case we should store the latest observed block.

akolotov:
OK. Let's try such kind of solution and test it carefully with high traffic on Ropsten or another PoW chain with big number of miners. The main idea is to test cases when soft forks appearing in the chain.

patitonar:
I'm closing this PR in favor of #26 that implements the solution of listening to events only on blocks that has the required block confirmations and is also based on the new architecture of the bridge.

PR #25, opened 2018-06-04
New architecture

### LINKED ISSUE #4: Performance optimization
Think about how we can increase performance

1. 
since there are 3 modules: one way would be run 3 separate processes: deposits, withdrawals, collectedSignatures
the only problem is shared nonce between process deposits and withdrawals

2. Think about what we can utilize for multicore workers
https://nodejs.org/api/cluster.html

patitonar:
This is a work in progress. 

The code base for the Watcher and Sender are implemented.
So far the `Deposit` event Watcher logic and `Home` Sender are implemented and configured.

To run the code:
- Install RabbitMQ and Redis
- Add env variables on `.env`. Example:
 `QUEUE_URL=amqp://user:password@127.0.0.1`
 `REDIS_URL=redis://:authpassword@127.0.0.1`
- Run Deposit Watcher: `npm run watcher:deposit`
- Run Home Sender: `npm run sender:home`

To interact with redis you can use [redis-cli](https://redis.io/topics/rediscli)
To watch rabbitMQ queues you can enable [Management Plugin](https://www.rabbitmq.com/management.html) and login on [http://localhost:15672](http://localhost:15672) 

Pending work:

- [x] Add collected signatures logic and config
- [x] Add withdraw logic and config
- [x] Add foreign sender config
- [x] Dev logs/Code Cleanup
- [x] Update Readme

Closes #4


patitonar:
Added missing events. For now, to run the watcher and senders:
- `npm run watcher:deposit`
- `npm run watcher:collected-signatures`
- `npm run watcher:withdraw`
- `npm run sender:home`
- `npm run sender:foreign`

src/utils/utils.js
```diff
@@ -0,0 +1,23 @@
+async function syncForEach(array, callback) {
+  for (let index = 0; index < array.length; index++) {
+    await callback(array[index], index, array)
+  }
+}
+
+function asyncForEach(array, callback) {
```
fvictorio: Using this seems to be the same as doing `array.map(callback)`. If that's the case, the abstraction is not worth it in my opinion.

.env.example
```diff
@@ -10,10 +10,13 @@ VALIDATOR_ADDRESS_PRIVATE_KEY=
 GAS_PRICE_SPEED_TYPE=standard
 GAS_PRICE_FALLBACK=1
 
+QUEUE_URL=amqp://user:password@127.0.0.1
+REDIS_URL=redis://:authpassword@127.0.0.1:6379/
```
fvictorio: Here it may be a good idea to use sensible defaults. For example:

```

File: src/services/amqpClient.js [hunk 3]
```diff
@@ -0,0 +1,50 @@
...
+
+connection.on('disconnect', () => {
+  console.error('Disconnected from amqp Broker')
+})
+
+function connectWatcherToQueue({ queueName, cb }) {
+  connection.createChannel({
+    setup(channel) {
+      const sendToQueue = data =>
+        channel.sendToQueue(queueName, Buffer.from(JSON.stringify(data)), { persistent: true })
```
fvictorio: I see that `node-amqp-connection-manager` has a `json` option that seems to automatically stringify and parse JSONs in the messages. Would it make sense to use it here?

src/watcher.js
```diff
@@ -0,0 +1,88 @@
...
+const path = require('path')
+const Web3 = require('web3')
+const { connectWatcherToQueue } = require('./services/amqpClient')
+const { getBlockNumber } = require('./utils/utils')
+const processDeposits = require('./events/processDeposits')
+const processCollectedSignatures = require('./events/processCollectedSignatures')
+const processWithdraw = require('./events/processWithdraw')
+const { redis } = require('./services/redisClient')
+
+const config = require(path.join('../config/', process.argv[2]))
```
fvictorio: Maybe we should add a check that `process.argv.length >= 3` here, with a proper error message.

src/watcher.js
```diff
@@ -0,0 +1,88 @@
...
+      return processWithdraw(events)
+    default:
+      return []
+  }
+}
+
+async function main({ sendToQueue }) {
+  try {
+    const lastBlockNumber = await getBlockNumber(web3Instance)
+    if (lastBlockNumber === undefined || lastBlockNumber === lastProcessedBlock) {
```
fvictorio: In what circumstances is `lastBlockNumber` undefined?

src/sender.js
```diff
@@ -0,0 +1,114 @@
...
+const config = require(path.join('../config/', process.argv[2]))
+
+const provider = new Web3.providers.HttpProvider(config.url)
+const web3Instance = new Web3(provider)
+const nonceLock = `lock:${config.id}:nonce`
+const nonceKey = `${config.id}:nonce`
+let chainId = 0
+
+async function initialize() {
+  chainId = await sendRawTx({
```
fvictorio: Here we can use `web3.eth.net.getId()`, unless there is a good reason to use raw HTTP.

src/sender.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+function updateNonce(nonce) {
+  redis.set(nonceKey, nonce)
+}
+
+async function main({ msg, ackMsg, nackMsg, sendToQueue }) {
+  if (redis.status !== 'ready') {
+    console.log('Redis not connected.')
+    await setTimeout(() => nackMsg(msg), 5000)
```
fvictorio: Two questions here:

1. Why the timeout instead of just immediately nacking the message?
2. Why the await? `setTimeout` doesn't return a promise.

src/services/amqpClient.js
patitonar: I think it make sense to use the `json` option. I'll give it a try and test it

src/watcher.js
patitonar: With this new code I think there is no use case for that `undefined` check, I will remove it.

src/sender.js
patitonar: You are right, I'll fix that line into `nackMsg(msg)`

src/services/amqpClient.js
patitonar: Tested and the `json` option automatically stringify data on `sendToQueue` but you still have to parse the data manually on the consumer. Anyway, it's an improvement so I'll add it.

rstormsf:
looks pretty good to me

rstormsf:
Plz specify in README.md which redis and rabbitmq versions are used/supported

PR #26, opened 2018-06-11
Listen events on confirmed blocks

### LINKED ISSUE #5: Transaction validation after specific number of confirmed block

patitonar:
This PR depends on #25 to be merged first.

This implementation let Watchers listen to events only on blocks that has the required block confirmations as explained [here](https://github.com/poanetwork/bridge-nodejs/pull/21#issuecomment-392924603). Performed tests on Kovan and Ropsten networks.

Closes #5


rstormsf:
Ok. I'm looking into it

akolotov:
@patitonar did you think about a test to make sure that transaction was returned back to transaction pool (become unconfirmed again) because a block with this transaction was considered as an uncle?
The idea to reproduce this scenario (we thought about this with Roman) is the following:
1. you need two nodes which are able to mine new blocks
2. at some moment you need to disable RLPx link between these two nodes as so they will not be able to sync.
3. you need to send a deposit transaction to one node and wait when it is verified and included in the blockchain. Make sure that bridge is communicating with this node by RPC. Stop the mining process on this node.
4. wait for 3-4 blocks on the second node and enable RLPx link between nodes again.
5. make sure that blocks mined on the the first node were obsoleted and all corresponding transactions were re-mined and included in new blocks

src/watcher.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+const config = require(path.join('../config/', process.argv[2]))
+
+const provider = new Web3.providers.HttpProvider(config.url)
+const web3Instance = new Web3(provider)
+const bridgeContract = new web3Instance.eth.Contract(config.abi, config.contractAddress)
+const lastBlockRedisKey = `${config.id}:lastProcessedBlock`
+let lastProcessedBlock = 0
+let requiredBlockConfirmations = 1
```
rstormsf: @patitonar 
can  we set it from .env file? HOME/FOREIGN_FALLBACK_BLOCK_CONFIRMATION

@akolotov agree?

rstormsf:
@patitonar I checked the code, it looks good, however it doesn't cover the case that @akolotov has described.

Basically  if I send a tx to blockchain and it gets mined. After 2-3 block the chain gets forked so the main chain no longer will have such transaction.
My proposal:
After RequiredConfirmation  has passed: ask blockchain again to make sure it's still  there.

src/watcher.js
patitonar: `let requiredBlockConfirmations = 1` works only as initialization, that value will be overwritten by `requiredBlockConfirmations()` call from the contract. Should we add `HOME/FOREIGN_FALLBACK_BLOCK_CONFIRMATION` on .env file that would be used if present instead of the value from the contract?

akolotov: It is assumed that we will always get number of block required for confirmation from the contract so there is no case when `HOME/FOREIGN_FALLBACK_BLOCK_CONFIRMATION` could be used. If we cannot invoke `requiredBlockConfirmations()` it means that something wrong with contracts and we should not continue operating.

In order to get rid of confusion with initialization of the  variable `requiredBlockConfirmations` does it make sense to initialize it as `null` in the line 24 to present that it is a variable declaration rather than a initialization of a value?

src/watcher.js
```diff
@@ -71,16 +74,20 @@ function processEvents(events) {
...
+  const lastBlockNumber = await getBlockNumber(web3Instance)
+  return lastBlockNumber - requiredBlockConfirmations + 1
+}
+
 async function main({ sendToQueue }) {
   try {
-    const lastBlockNumber = await getBlockNumber(web3Instance)
-    if (lastBlockNumber === lastProcessedBlock) {
+    const lastBlockToProcess = await getLastBlockToProcess()
+    if (lastBlockToProcess === lastProcessedBlock) {
```
akolotov: Now `lastBlockToProcess` could be less than `0` if the bridge is run on a just created testnet (when `requiredBlockConfirmations` is greater than a latest block number).
Will `bridgeContract.getPastEvents(..)` handle this properly?

src/watcher.js
```diff
@@ -71,16 +74,20 @@ function processEvents(events) {
   }
 }
 
+async function getLastBlockToProcess() {
+  const lastBlockNumber = await getBlockNumber(web3Instance)
+  return lastBlockNumber - requiredBlockConfirmations + 1
```
akolotov: Is this formula correct? Imagine that we have configured `0` confirmations which means that the bridge will handle an event as soon as a transaction was included in the block. In this case the formula will get `lastBlockToProcess` as the latest block number plus `1` - this means that `getPastEvents` needs to check events from the block which does not exist yet. Please correct me if I am wrong.

patitonar: Thanks for pointing this out @akolotov. I assumed that the minimum value for `requiredBlockConfirmations` would be `1` (just the block in which the Tx was mined). But if it can be `0` then I need to fix the formula. Maybe @rstormsf can clarify this

src/watcher.js
patitonar: I'm not sure `bridgeContract.getPastEvents(..)` will handle this properly, I'll add a fix to prevent getting the events if `lastBlockToProcess` is less than `lastProcessedBlock`

patitonar:
@akolotov I'll work on a fix to allow the bridge to use the proper updated `requiredBlockConfirmations` value without the need to restart.

src/watcher.js
rstormsf: Alex is right. Someone can say requiredblock is 0

PR #27, opened 2018-06-12
Add docker and e2e test

patitonar:
This PR depends on #25 to be merged first.

Add docker configuration to run the app and run the e2e tests in Travis with the new architecture of the bridge.


fvictorio:
@rstormsf @akolotov Can we merge this? (I just noticed that we didn't add you as reviewers, sorry about that.)

akolotov:
I have added Pavel to be familiar the suggested approach

rstormsf:
@patitonar 
Please add to gitignore all  files that expose private keys.
Reason: as a developer, when I add  my own keys, and do some improvement,  I could accidentally add files to git.

rstormsf:
instead, just add to git .file.example.env and add to documentation to copy those files and fill in  the values

phahulin:
Few suggestions:
1. it's probably best to separate `.env` into separate files
- `public.env` that would store configuration shared by all validator nodes and will be committed to git
- actual `.env` that would store validator private key and will not be committed to git
- additional env file for tests only

2. can `GAS_PRICE_SPEED_TYPE` be renamed to `GAS_PRICE_SPEED` to match `poa-bridge`?

3. if one of workers inside docker container crashes (for whatever reason), is there a way to automatically restart it? Is it possible to restart with a small delay (1-2 sec)?

4. there's an extra `\` in README.md (invisible in github)
```
...
Make sure your HOME\_MIN\_AMOUNT\_PER\_TX is same as in your .env deployment contract
...
```

fvictorio:
@phahulin I think that's a good idea for the bridge, not so for the e2e (see my [comment](https://github.com/poanetwork/bridge-nodejs/pull/29#issuecomment-398384332) in the other PR). Separating the environment files in public and private would make it easier to get the bridge working (both manually and with docker) and it would work as a sort of documentation.

Do you mind if I fill this as a separate issue and do it in another PR? Since this one has a dependent PR, it's a little messy to add more commits here.

phahulin:
@fvictorio I see, this PR is only for setting up tests. Then I agree with you, it should addressed in a separate issue.

akolotov (approved):

phahulin (approved):
Please address suggestions in new separate issues.

fvictorio:
@phahulin Sorry for not responding to the other points before.

> it's probably best to separate .env into files

Created issue #32.

> can GAS_PRICE_SPEED_TYPE be renamed to GAS_PRICE_SPEED to match poa-bridge?

Created issue #33.

> if one of workers inside docker container crashes (for whatever reason), is there a way to automatically restart it? Is it possible to restart with a small delay (1-2 sec)?

We can do that with docker (it has a restart option) or we can use a process manager. That will depend on how we deploy this.

> there's an extra \ in README.md (invisible in github)

I didn't understand this. As far as I know, those `\_` are necessary to avoid rendering the text as italics.

phahulin:
@fvictorio Thanks for the issues

> We can do that with docker (it has a restart option) or we can use a process manager. That will depend on how we deploy this.

I think to do that in docker is better, because it will make the whole setup less dependent on the external conditions like OS, specific process manager etc. Process manager can be used to ensure that docker daemon itself is running

> As far as I know, those \_ are necessary to avoid rendering the text as italics.

Yes, sorry. I mistakenly thought it was inside the `code quotes`

rstormsf:
Question: is there a change in README file that needs to address how to run those tests and if any external dependencies need to be installed?

Chewypewy (commented):
Does this help

PR #29, opened 2018-06-18
Improve e2e tests

fvictorio:
Depends on #27.

- Use mocha for running the tests.
- Use [promise-retry](https://www.npmjs.com/package/promise-retry) to check until the balance changes.
- Add scenario for testing foreign-to-home transaction.

I also added a `constants.json` file to (eventually) have a single source of truth for addresses, private keys, etc. But right now we still have a lot of duplication. I didn't find a simple way to do this, because it will mean generating the `.env` files and I want to avoid that complexity for the time being.


rstormsf:
same issue as in #27 
Private keys when changed, git system will want to add them on change. those files should be ignored and provided instruction how to supply private keys.

fvictorio:
These private keys are for the hardcoded addresses in the private chain spec. I'm not sure I see in which scenario someone would add their own PKs. I'd like to have all the addresses and private keys used in the test in a single file, but even then I think I would commit it.

I think of this as a hardcoded password in a e2e test for some login, for example. You would create a user and then test the login with the correct password. In that scenario, you wouldn't make anything to hide that password in the code, because that user doesn't exist, it's just for a test. Am I missing something here?

PR #31, opened 2018-06-21
Avoid using HTTP RPC URLs if it's not explicitly specified

### LINKED ISSUE #10: HTTPS must be forced for RPC connections
As per recommendation from a team provided security audit for POA bridge it is needed to force https connection for RPC communications.

In other words RPC connection must not succeed if HTTP is used and the bridge instance must stop.

In order to simplify configuration of testing environment and to allow to use plain HTTP requests, a new parameter like `FORCE_HTTPS` should be introduced in the configuration file. It's value should be considered as `yes` by default (even if the parameter is not pointed out in the configuration file). So, If a developer would like to use plain HTTP connection instead of HTTPS, the parameter needs to be set to `no`.

fernandomg:
Closes #10 

A new `env` variable is introduced: `ALLOW_HTTP`. If it's not explicitly set to `yes` process will fail if there's an RPC_URL pointing to an `http` address.


.env.example
```diff
@@ -1,3 +1,4 @@
+#ALLOW_HTTP=yes
```
fvictorio: Personal preference: I'd prefer to have some kind of default value here (`ALLOW_HTTP=no`) instead of having a commented line. Specially if use add `dotenv-safe` eventually.

src/utils/utils.js
```diff
@@ -4,6 +4,19 @@ async function syncForEach(array, callback) {
   }
 }
 
+function checkHTTPAllowance(urls) {
+  urls.forEach(url => {
+    if (!/^https.*/.test(url)) {
+      if (process.env.ALLOW_HTTP !== 'yes') {
```
fvictorio: I don't think that the method should be responsible of checking the environment variable (nor throwing, for that matter). I'd move most of this logic to the caller and leave just a `isHTTPS(url)` function here.

.env.example
fernandomg: In that case it'll be better to add extra info in `README.md` to explain the expected values for it.

Despite it looks obvious that the opposite to `no` is be `yes`, I prefer avoid leaving place to misunderstandings.

fernandomg: done

src/utils/utils.js
fernandomg: After debating about the implementation and implications. `checkHTTPAllowance` was renamed to `checkHTTPS`, and changed its implementation to receive the `env` variable as a parameter, as well as it checks for only one URL at a time.

rstormsf:
@patitonar merge conflicts

patitonar:
@rstormsf merge conflicts solved

rstormsf:
is  it ok that CI fails?

PR #38, opened 2018-06-25
Update e2e test for block confirmations

patitonar:
Update e2e test and docker-compose config to support changes from #26 #31 and #36

### DIFF
--- a/e2e/test/transactions.js
+++ b/e2e/test/transactions.js
@@ -3,7 +3,7 @@ const Web3 = require('web3')
 const assert = require('assert')
 const promiseRetry = require('promise-retry')
 const { user } = require('../constants.json')
-const { generateNewBlockWithTx } = require('../utils/utils')
+const { generateNewBlock } = require('../utils/utils')
 
 const abisDir = path.join(__dirname, '..', 'submodules/poa-bridge-contracts/build/contracts')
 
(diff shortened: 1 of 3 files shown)


e2e/test/transactions.js
```diff
@@ -27,14 +28,34 @@ describe('transactions', () => {
...
+      async retry => {
+        const lastBlockNumber = await homeWeb3.eth.getBlockNumber()
+        if (lastBlockNumber === depositTx.blockNumber + 2) {
+          await generateNewBlockWithTx(homeWeb3, user.address)
+        } else {
+          retry()
+        }
+      },
+      {
+        retries: 1000000,
```
fvictorio: You can use `forever: true` instead of doing this. It's not explicitly documented in `promise-retry`, but since it just sends the options to [retry](https://github.com/tim-kos/node-retry), you can use whichever options are there.

e2e/test/transactions.js
```diff
@@ -52,12 +73,15 @@ describe('transactions', () => {
...
         from: user.address,
-        gasLimit: '1000000'
+        gas: '1000000'
       })
       .catch(e => {
         console.error(e)
       })
 
+    // Send a second Tx to generate a new block
+    await generateNewBlockWithTx(foreignWeb3, user.address)
```
fvictorio: Personal preference: I would call it just `generateNewBlock`.

e2e/test/transactions.js
```diff
@@ -27,14 +28,34 @@ describe('transactions', () => {
...
       gasPrice: '1',
-      gasLimit: '50000',
+      gas: '50000',
       value: '1000000000000000000'
     })
 
+    // Send a Tx to generate a new block
+    await generateNewBlockWithTx(homeWeb3, user.address)
+
+    // wait to send a Tx to generate a new block
```
fvictorio: What we are doing here is a little tricky, so a more verbose comment would help.

We send a transaction that will generate a Deposit event, and then a trivial transaction to generate a new block (since the watcher is configured to wait 1 block). After this, the bridge should create a new transaction with a CollectedSignatures event and, for the same reason, we generate another trivial transaction so that the second watcher gets this.

e2e/test/transactions.js
```diff
@@ -27,14 +28,34 @@ describe('transactions', () => {
...
+    await promiseRetry(
+      async retry => {
+        const lastBlockNumber = await homeWeb3.eth.getBlockNumber()
+        if (lastBlockNumber === depositTx.blockNumber + 2) {
```
fvictorio: I would use `>=` here. It's very unlikely for this to be necessary, but it won't hurt and it's a little more defensive.

PR #39, opened 2018-06-27
Handle insufficient funds

### LINKED ISSUE #13: Incorrect handling of transactions in case of insufficient funds
If a validator has no enough funds to handle transaction the bridge is trying to send transaction, fails and updating the database:

```
Found 1 Deposits on Home Network
Error! { code: -32010,
  message: 'Insufficient funds. The account you tried to send transaction from does not have enough funds. Required 2314865000000000 and got: 0.' }
1 '# processing deposit' '0xa85bb9dca943978bff7054db8172cc7c382cb321364e3b7fd3f137d30bb51983' undefined
writing db deposits 91
```

It leads to necessity to stop the bridge, roll back the changes in database manually and start the bride again after the validator balance refill.

So, in order to handle 'insufficient funds' issue properly the bridge should:
1. check balance of validator on Home and Foreign side as soon as the corresponding event (`Deposit`, `CollectedSignaturs` or `Withdraw`) to relay transaction is received
2. only if the balance is positive, relay transaction
3. otherwise the bridge must stop operations in the corresponding direction until the validator balance is refilled. **No database updates must be done** for the corresponding event type.

fvictorio:
Closes #13.

If some transaction fails because of insufficient funds, close the connection and don't wait for the balance of the validator to change.

To avoid weird circumstances, we added a `prefetch(1)` that means that only one message will be processed at the same time. This shouldn't affect the performance significantly since most of the processing is locked by the nonce anyway. In fact, this may render the lock unnecessary, but since it won't hurt either we left it. We should consider removing it, though.

### DIFF
--- a/src/sender.js
+++ b/src/sender.js
@@ -55,7 +55,7 @@ function updateNonce(nonce) {
   return redis.set(nonceKey, nonce)
 }
 
-async function main({ msg, ackMsg, nackMsg, sendToQueue, close }) {
+async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
   try {
     if (redis.status !== 'ready') {
       console.log('Redis not connected.')
@@ -77,6 +77,7 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, close }) {
     console.log('Nonce Locked! After: ', timeWaitingForLock)
 
     let nonce = await readNonce()
+    let insufficientFunds = false
     const failedTx = []
 
     await syncForEach(txArray, async job => {
@@ -99,19 +100,16 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, close }) {
       } catch (e) {
         console.error(e)
         console.error(`Tx Failed for event Tx ${job.transactionReference}`)
+        failedTx.push(job)
+
         if (e.message.includes('Insufficient funds')) {
           console.error('Insufficient funds')
-          waitForFunds(web3Instance, () => initialize())
-          close()
-        } else {
-          failedTx.push(job)
-
-          if (
-            e.message.includes('Transaction nonce is too low') ||
-            e.message.includes('transaction with same nonce in the queue')
-          ) {
-            nonce = await readNonce(true)
-          }
+          insufficientFunds = true
+        } else if (
+          e.message.includes('Transaction nonce is too low') ||
+          e.message.includes('transaction with same nonce in the queue')
+        ) {
+          nonce = await readNonce(true)
         }
       }
     })
@@ -127,6 +125,11 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, close }) {
       await sendToQueue(failedTx)
     }
     ackMsg(msg)
+
+    if (insufficientFunds) {
+      channel.close()
+      waitForFunds(web3Instance, VALIDATOR_ADDRESS, initialize)
+    }
   } catch (e) {
     console.error(e)
     nackMsg(msg)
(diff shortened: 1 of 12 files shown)


src/sender.js
```diff
@@ -121,6 +125,11 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue }) {
       await sendToQueue(failedTx)
     }
     ackMsg(msg)
+
+    if (insufficientFunds) {
+      channel.close()
+      waitForFunds(web3Instance, VALIDATOR_ADDRESS, initialize)
```
patitonar: I think it would be useful to add 2 logs. One with a message that it will stop processing jobs because of the funds, and a second log with a message that the validator have funds again and it will start processing jobs

patitonar: I think it would be useful to add 2 logs. One with a message that it will stop processing jobs because of the funds, and a second log with a message that the validator have funds again and it will start processing jobs

fvictorio: Added.

src/utils/utils.js
```diff
@@ -16,7 +18,27 @@ function checkHTTPS(ALLOW_HTTP) {
   }
 }
 
+async function waitForFunds(web3, address, cb) {
+  const balance = web3.utils.toBN(await web3.eth.getBalance(address))
+
+  promiseRetry(
+    async retry => {
+      const newBalance = web3.utils.toBN(await web3.eth.getBalance(address))
+      if (newBalance.eq(balance)) {
```
akolotov: you are looking for any difference in balances. Does it make sense to resume only if balance is enough to send particular transaction (gas usage * gas price)?

src/sender.js
```diff
@@ -101,7 +107,10 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue }) {
         console.error(`Tx Failed for event Tx ${job.transactionReference}`)
         failedTx.push(job)
 
-        if (
+        if (e.message.includes('Insufficient funds')) {
+          console.error('Insufficient funds')
```
akolotov: Will the similar message be logged anyway in the line 106? Will we have duplicated lines in the logs?

fvictorio: You are right. We can just remove this comment, or make it more clear. What about something like:

```Insufficient funds: ${currentBalance}. Stop consuming messages until the balance is ${minimumBalance}.```

File: src/utils/utils.js (hunk 2 above)
Comment (fvictorio) [reply src/utils/utils.js]: Fixed, thanks.

File: src/sender.js [hunk 4]
```diff
@@ -101,7 +110,16 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue }) {
         console.error(`Tx Failed for event Tx ${job.transactionReference}`)
         failedTx.push(job)
 
-        if (
+        if (e.message.includes('Insufficient funds')) {
+          insufficientFunds = true
+          const currentBalance = await web3Instance.eth.getBalance(VALIDATOR_ADDRESS)
+          minimumBalance = toBN(job.gasEstimate)
+            .add(toBN(200000))
```
akolotov: Could you clarify why you added 200_000 of gas? Does it make sense to add 20%?
Whatever the value you are using, let use a constant and provide a comment to expose its meaning. Thanks

fvictorio: Oh, yes, sorry. That is the amount that is being added some lines above to the estimated gas. But I totally agree that it should be in a constant. Maybe something like `EXTRA_GAS = 200000`? Or we could use a percentage.

akolotov: I would prefer to use a percentage. Otherwise for transactions which requires less than 100 thousands the gas parameter will have the value 3 times more than required.

fvictorio: Done. I added a constant for the extra percentage and a function that applies this increase.

rstormsf:
I always  add a little more gas, cause estimateGas can be wrong a little bit depending which  node is executing the  tx: geth or parity.

fvictorio:
@rstormsf Do you think that adding a 5% to the estimated gas should be OK?

rstormsf:
@fvictorio lets add 25% to the  gas.

fvictorio:
@rstormsf Done.

PR #43, opened 2018-07-04
Improve gas price management

### LINKED ISSUE #6: Improvements in gas price management
At this moment the code is written as so:
1. try to get gas price from the oracle
https://github.com/poanetwork/bridge-nodejs/blob/a23ecb78572f3e1928d9129eff8d4118b2618b57/src/gasPrice.js#L9
2. if the oracle is not accessible, the value from configuration file is used
https://github.com/poanetwork/bridge-nodejs/blob/a23ecb78572f3e1928d9129eff8d4118b2618b57/src/gasPrice.js#L14

There are few issues here:
1. The same oracle is used for both sides: Home and Foreign although it is assumed that these are two different networks and gas price for them could be different.
2. The fallback value is the same for both networks.

In order to address these issues the following is suggested to be implemented as result of previous discussion (https://github.com/poanetwork/poa-bridge/issues/16#issuecomment-386784045):

1. The following options to allow configure a JSON source of gas price should be available in the configuration file:
```
HOME_GAS_PRICE_ORACLE_URL = "https://gasprice.poa.network/"
HOME_GAS_PRICE_SPEED_TYPE = "fast"
HOME_GAS_PRICE_UPDATE_TIMEOUT = 600
FOREIGN_GAS_PRICE_ORACLE_URL = "https://gasprice.poa.network/"
FOREIGN_GAS_PRICE_SPEED_TYPE = "fast"
FOREIGN_HOME_GAS_PRICE_UPDATE_TIMEOUT = 600
```
Expecting that the gas price oracle will return information in the following form:
```
{"block_number":5559089,"block_time":14.824,"instant":50.0,"standard":4.0,"slow":2.0,"fast":5.0,"health":true}
```
And `HOME_GAS_PRICE_UPDATE_TIMEOUT` and `FOREIGN_HOME_GAS_PRICE_UPDATE_TIMEOUT` mean how often the corresponding oracle is requested to get updates.

2. If a request to the corresponding oracle is failed, the gas price needs to be requested from the corresponding contract by calling `gasPrice()` (https://github.com/poanetwork/poa-bridge-contracts/blob/93969cd934190649d42691f054676d0d16a9777e/contracts/upgradeable_contracts/U_BasicBridge.sol#L29) 

3. If the value from contract cannot be obtained the latest known value of gas price should be used.

4. Gas price obtained by either one of the way presented in 1, 2 or 3 will be considered as the cached gas price value.

4. In order to reduce number of requests to the oracles, transactions should be constructed with cached gas price value instead of sending request to the oracle for every transaction.

fvictorio:
Closes #6.

This PR improves the way the gas price is obtained:

- Every some seconds (configurable), the gas price is obtained. This is done by fetching the oracle, calling the contract method or using the fallback price.
- This value is saved in memory and, when asked, this cached value is returned.

There are several changes that I made to be able to test stuff more easily:

- Added unit tests. Since we already had a `tests` directory, I renamed it to `scripts` and created a `test` directory for the unit tests.
- These scripts were getting the numnber of deposits/withdrawals to create from the environment. I think it's easier to pass this as an argument to the command, so I modified them to do this.

Finally, I modified the Travis configuration to run the linter and the unit tests before running the e2e tests.


fvictorio:
I just noticed that I may have done something wrong: when the oracle and the contract fail, the cached value is updated with the fallback value, but maybe the fallback value should be used just as the initial value (so that is used if the oracle and the contracts fail every time).

Another thing I don't like is that the fallback value will be used until the `setInterval` is executed the first time. So if this function is executed every 1 minute, during the first minute the fallback gas price will be used.

Both things are easy to fix, so let me know if you think they should be changed.

src/services/gasPrice.js
```diff
@@ -1,19 +1,99 @@
...
+      console.error(`There was a problem getting the gas price from the contract. ${e.message}`)
+    }
+  }
+  return gasPrice
+}
+
+let homeGasPriceInterval = null
+let foreignGasPriceInterval = null
+
+async function start() {
```
patitonar: I think the sender needs the gas price for only one chain, Home or Foreign.

fvictorio:
Fixed @patitonar's comment and also modified the function that fetches the gas price so that it returns null if both the oracle and the contract fail. If that function returns null, the last cached gas price is used. This value is initialized with the fallback value, so the resulting behavior is:

- Start with the fallback value
- Fetch the gas price (success): the gas price is updated
- Fetch the gas price (fail): the last fetched value is used.

PR #47, opened 2018-07-11
Optimization for blocks filter on first run

### LINKED ISSUE #7: Optimization for blocks filter out used at very first bridge run
At the current implementation the number of blocks which needs to be used to look for bridge events could be several millions even if the bridge contracts were deployed recently:

https://github.com/poanetwork/bridge-nodejs/blob/a23ecb78572f3e1928d9129eff8d4118b2618b57/src/processCollectedSignatures.js#L51

`db.processedBlock` is taken from the bridge database and equals `0` during very first start.

At the same time every bridge contract contain the number of block which contract deployment transaction was included in:
* Home: https://github.com/poanetwork/poa-bridge-contracts/blob/3b33b2855f89849bce5a49fd417669468c665d41/contracts/upgradeable_contracts/U_HomeBridge.sol#L31
* Foreign: https://github.com/poanetwork/poa-bridge-contracts/blob/3b33b2855f89849bce5a49fd417669468c665d41/contracts/upgradeable_contracts/U_ForeignBridge.sol#L43

This block number could be used as the initial position to look for bridge event.

It is necessary to change the database initialization process:
https://github.com/poanetwork/bridge-nodejs/blob/a23ecb78572f3e1928d9129eff8d4118b2618b57/src/processDeposits.js#L24-L25
https://github.com/poanetwork/bridge-nodejs/blob/a23ecb78572f3e1928d9129eff8d4118b2618b57/src/processCollectedSignatures.js#L32-L33
https://github.com/poanetwork/bridge-nodejs/blob/a23ecb78572f3e1928d9129eff8d4118b2618b57/src/processWithdraw.js#L32-L33
as so if the initial value is zero the method `deployedAtBlock()` of the corresponding contract needs to be called.

patitonar:
Closes #7 

The Watchers will look for the following `.env` variables
```
HOME_START_BLOCK
FOREIGN_START_BLOCK
```

Added a script `scripts/getValidatorStartBlocks.js`. The script will return an array containing two elements as follows: `[homeStartBlock, foreignStartBlock]`

Examples on how to call the script:

- bash example
```
val=$(node scripts/getValidatorStartBlocks.js)
echo "Blocks returned: $val"
```
- node example
```
const getValidatorBlocks = require('./scripts/getValidatorStartBlocks')
getValidatorBlocks().then(result => console.log('Blocks returned:', result))
```

```
// Output 
Blocks returned: [ 2626974, 7428613 ]
```


scripts/getValidatorStartBlocks.js
```diff
@@ -0,0 +1,46 @@
...
+    const deployedAtBlock = await bridgeContract.methods.deployedAtBlock().call()
+
+    const validatorContractAddress = await bridgeContract.methods.validatorContract().call()
+    const validatorContract = new web3Instance.eth.Contract(
+      bridgeValidatorsABI,
+      validatorContractAddress
+    )
+
+    const validatorAddedEvents = await validatorContract.getPastEvents('ValidatorAdded', {
+      fromBlock: 0,
```
phahulin: Maybe better use `deployedAtBlock` instead of `0`?

scripts/getValidatorStartBlocks.js
```diff
@@ -0,0 +1,46 @@
...
+    return 0
+  }
+}
+
+async function main() {
+  const { HOME_RPC_URL, FOREIGN_RPC_URL, HOME_BRIDGE_ADDRESS, FOREIGN_BRIDGE_ADDRESS } = process.env
+
+  const homeStartBlock = await getStartBlock(HOME_RPC_URL, HOME_BRIDGE_ADDRESS, homeABI)
+  const foreignStartBlock = await getStartBlock(FOREIGN_RPC_URL, FOREIGN_BRIDGE_ADDRESS, foreignABI)
+  console.log([homeStartBlock, foreignStartBlock])
```
phahulin: could you use an object `[homeStartBlock, foreignStartBlock]` -> `{homeStartBlock, foreignStartBlock}` here and in `return` statement?

patitonar:
@phahulin updated returned value of script. Now it is an object.
 Output Example: `{ homeStartBlock: 2626974, foreignStartBlock: 7428613 }`

phahulin:
@patitonar thanks, but I'll have to ask you to additionally wrap console output like this
```
console.log(JSON.stringify(result))
```
otherwise it's not parsed correctly by ansible.
Sorry about that.

patitonar:
@phahulin Added `JSON.stringify` to the console output. 
Example ouput: `{"homeStartBlock":2626974,"foreignStartBlock":7428613}`

scripts/getValidatorStartBlocks.js
akolotov: Do you think it makes sense to extend functionality of Validator contract to have `deployedAtBlock` for this contract as well?

patitonar: @akolotov I think having `deployedAtBlock` on Validator contract will be useful for this case to avoid listening from block 0. If you agree with this, should I make a PR in the poa-bridge-contracts repo to add `deployedAtBlock` ?

patitonar: @akolotov Added a PR for this [https://github.com/poanetwork/poa-bridge-contracts/pull/54](https://github.com/poanetwork/poa-bridge-contracts/pull/54). Once it is merged, I can update this PR to use `deployedAtBlock` from BridgeValidators contract

akolotov: @patitonar I have approved https://github.com/poanetwork/poa-bridge-contracts/pull/54

rstormsf: @patitonar merged

patitonar:
@rstormsf Script updated to use `deployedAtBlock` from Validator contract

PR #57, opened 2018-07-27
Add reset last block script

### LINKED ISSUE #46: Add a script to reset lastProcessedBlock
In some cases a bridge node may need to re-process past events (e.g. there was an error and some events were skipped). In the current (rust) version of the bridge it requires changing a text file that stores last processed block number, which is pretty straight-forward.

In this version of the bridge, block numbers are stored in redis, so changing them would require starting redis cli and running a command similar to `SET deposit:lastProcessedBlock 1234`, which might not be very convenient for validators.

Please add a single shell script that would automate this, with a minimal set of arguments, something like
```
./reset-lastBlock deposit 1234
```

patitonar:
Closes #46 

Examples of running the shell script for each watcher
```
./reset-lastBlock.sh signature-request 8149353
./reset-lastBlock.sh collected-signatures 8149353
./reset-lastBlock.sh affirmation-request 9372461
```

### DIFF
--- a/scripts/resetLastBlock.js
+++ b/scripts/resetLastBlock.js
@@ -1,0 +1,60 @@
+const Redis = require('ioredis')
+const path = require('path')
+require('dotenv').config({
+  path: path.join(__dirname, '../.env')
+})
+
+const redis = new Redis(process.env.REDIS_URL)
+
+redis.on('error', () => {
+  logError('Error: Cannot connect to redis')
+})
+
+if (process.argv.length < 4) {
+  logError(
+    'Please provide process key and new block value. Example:' +
+      '\n  signature-request 12345 ' +
+      '\n  collected-signatures 12345 ' +
+      '\n  affirmation-request 12345'
+  )
+}
+
+function logError(message) {
+  console.log(message)
+  process.exit(1)
+}
+
+function getRedisKey(name) {
+  const isErcToErc = process.env.BRIDGE_MODE && process.env.BRIDGE_MODE === 'ERC_TO_ERC'
+  const prefix = isErcToErc ? 'erc-' : ''
+  return `${prefix}${name}:lastProcessedBlock`
+}
+
+async function main() {
+  try {
+    const processName = process.argv[2]
+    const newBlockValue = process.argv[3]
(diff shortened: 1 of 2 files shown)


scripts/resetLastBlock.js
```diff
@@ -0,0 +1,60 @@
...
+    if (!value) {
+      logError(
+        'Error: Process key not found on redis. Please provide one of the following:' +
+          '\n  signature-request' +
+          '\n  collected-signatures' +
+          '\n  affirmation-request'
+      )
+    }
+
+    await redis.set(lastBlockRedisKey, newBlockValue)
```
phahulin: If `newBlockValue` should be an integer number, I suggest to add a check and explicit conversion here, something like
```js
newBlockValue = Number(newBlockValue);
if ( !Number.isInteger(newBlockValue) ) {
    logError('Expecting new block value to be an integer!' );
}
await redis.set(lastBlockRedisKey, newBlockValue);
```

scripts/resetLastBlock.js
```diff
@@ -0,0 +1,60 @@
...
+
+async function main() {
+  try {
+    const processName = process.argv[2]
+    const newBlockValue = process.argv[3]
+    const lastBlockRedisKey = getRedisKey(processName)
+
+    const value = await redis.get(lastBlockRedisKey)
+
+    if (!value) {
```
fvictorio: Notice that this won't allow setting a value for a key that hasn't been initialized yet.

scripts/resetLastBlock.js
fvictorio: If we do this, I think it should be done before (right after getting the value from `process.argv`).

scripts/resetLastBlock.js
patitonar: I think it's ok in this case. The script is intended for updating the existing values to re-process past events. On the first run, when keys are not initialized yet, the watcher will use `HOME_START_BLOCK` and `FOREIGN_START_BLOCK` from `.env`

scripts/resetLastBlock.js
patitonar: Done

akolotov:
@patitonar can we merge it?

PR #58, opened 2018-07-30
Extend message with contract address

### LINKED ISSUE #55: Extend signed message with Foreign bridge contract address
As per security audit held for Bridge smart contracts it was suggested to eliminate possibility of attack when validators approvals sent for `NATIVE-TO-ERC` bridge could be used for `ERC-TO-ERC` bridge.

In order to achieve this the message used to relay transactions to Foreign network was extended with an additional field which contains the address of Foreign bridge contract. Since it is assumed that the bridge with the same contract address is responsible for transferring specific asset only, the proposed changes are enough to address the comment apeeared during the audit.

The issue for `bridge-nodejs` was created to implement correct message composing by every validator prior to sending this message to Home brdige contract as part of `submitSignature` invokation.

Proposed changes for `processSignatureRequests.js`:
Instead of 
https://github.com/poanetwork/bridge-nodejs/blob/afa4b358476713aac2b90bbd1240c3d5f4ef0944/src/events/processSignatureRequests.js#L24-L28 
the following code will take place:
```
 const message = createMessage({ 
   recipient, 
   value, 
   transactionHash: signatureRequest.transactionHash,
   config.foreignBridgeAddress 
 })
```
assuming that the corresponding changes are made in `createMessage` of `utils/message.js`

The code of contracts which could be used to test proposed changes is being reviewed under https://github.com/poanetwork/poa-bridge-contracts/pull/57

fvictorio:
Closes #55.

This PR adds the bridge address to the messages created by the watchers.

I also used this PR to add a submodule with the `poa-bridge-contracts` repository (actually I moved it from the `e2e` directory to the root directory, since it was already there and it wasn't being used by the tests). This is a first step towards making easier to keep the bridge and the contracts in sync, and it's also useful as a way of documenting the dependency. There are some other things we could do here:

- Delete the `abis` directory and use the `build/contracts` directory in the submodule.
- Right now, the `Dockerfile` in the `e2e` directory needs to be updated too with the proper branch/commit of the contracts repository. It would be nice to remove this duplication.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 node_modules
-e2e/submodules
+submodules
+coverage
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "e2e/submodules/poa-bridge-contracts"]
-	path = e2e/submodules/poa-bridge-contracts
+	path = submodules/poa-bridge-contracts
 	url = https://github.com/poanetwork/poa-bridge-contracts.git
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -6,11 +6,11 @@ WORKDIR /stuff
 COPY package.json .
 COPY package-lock.json .
 
-RUN git clone https://github.com/poanetwork/poa-bridge-contracts.git
+RUN git clone https://github.com/rstormsf/poa-parity-bridge-contracts.git
 RUN mkdir submodules && \
-  mv poa-bridge-contracts submodules && \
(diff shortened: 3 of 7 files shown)


fvictorio:
Please don't merge this until https://github.com/poanetwork/poa-bridge-contracts/pull/57 is merged.

fvictorio:
This is ready for review. If you want to run the e2e tets locally, you may need to do `docker-compose build --no-cache` first, since otherwise it will not use the last commit in `refactor_v1`.

PR #59, opened 2018-08-02
Support several RPC URLs

### LINKED ISSUE #28: Ability to configure several RPC urls
At this moment it is possible to configure on one RPC url for each bridge direction:

```
HOME_RPC_URL=https://kovan.infura.io/mew
FOREIGN_RPC_URL=https://kovan.infura.io/mew
```

It means that if the remote node will not be accessible (e.g. the link is down or the node was shut down) the bridge will not be able get events and send transactions by using the corresponding RPC channel.

In order to increase availability of the bridge opearations it is suggested to introduce abiity to configure list of RPC URLs:

```
HOME_RPC_URL="https://sokol.poa.network/, https://some.rpc.node.at.home/"
FOREIGN_RPC_URL=https://kovan.infura.io/mew, https://some.rpc.node.at.foreign/
```

In this case the bridge should switch to the next RPC link in the list if the previous URL in the list is not accessible. As soon as the the link is switched the corresponding log message should appear, e.g.

```
The RPC channel (<url is here>) is not accessible. Swithcing to the next RPC link (<url is here>)
```

If only one RPC link is configured the bridge should use a random delay (e.g. from 5 to 30 seconds) to try to reconnect to the RPC link.

If the end of the list is reached the first URL from the list must be checked again and the bridge should switch to this channel.

If all channels are not available the bridge should use a random delay (e.g. from 5 to 30 seconds) to try to reconnect to the first RPC link from the list.

fvictorio:
Closes #28.

This PR allows configuring several URLs in the `HOME_RPC_URL` and `FOREIGN_RPC_URL` environment variables.

To do this, two things were necessary: creating a new web3 provider that accepts several URLs (see `HttpListProvider`) and creating a `RpcUrlsManager` service that, given a function, will try it with each available URL until one of them works. This service also centralizes the access to the URLs, instead of having `process.env.HOME_RPC_URL`s all over the code.

---

To test this, I created a simple express app that acts as a proxy and logs each request. Then I started two proxy apps that redirect everything to the port `8545` and two other that redirect to the port `8546`, where I had two parity nodes running. Then I configured the bridge to use two of the proxies as home URLs and the other two as foreign URLs. Finally, I killed and restarted the proxies in several combinations while sending transactions to check that everything was working.

The code for this express proxy is:

```
const proxy = require('express-http-proxy')
const app = require('express')()

const { PORT_FROM, PORT_TO } = process.env

app.use('/', proxy(() => {
  console.log('Request')
  return `http://localhost:${PORT_TO}`
}))

app.listen(PORT_FROM)
```

Then, in four different terminals, I did:

```
PORT_FROM=8555 PORT_TO=8545 node index.js
PORT_FROM=8565 PORT_TO=8545 node index.js
PORT_FROM=8556 PORT_TO=8546 node index.js
PORT_FROM=8566 PORT_TO=8546 node index.js
```

And in the `.env` file I used:

```
HOME_RPC_URL=http://localhost:8555,http://localhost:8565
FOREIGN_RPC_URL=http://localhost:8556,http://localhost:8566
```

### DIFF
--- a/src/utils/HttpListProvider.js
+++ b/src/utils/HttpListProvider.js
@@ -1,0 +1,35 @@
+const fetch = require('node-fetch')
+
+function HttpListProvider(urls) {
+  if (!urls || !urls.length) {
+    throw new Error(`Invalid URLs: '${urls}'`)
+  }
+
+  this.urls = urls
+  this.currentIndex = 0
+}
+
+HttpListProvider.prototype.send = async function(payload, callback, retries = 0) {
+  if (retries === this.urls.length) {
+    callback(new Error('Request failed for all urls'))
+  }
+
+  const url = this.urls[this.currentIndex]
+
+  try {
+    const result = await fetch(url, {
+      headers: {
+        'Content-type': 'application/json'
+      },
+      method: 'POST',
+      body: JSON.stringify(payload)
+    }).then(request => request.json())
+
+    callback(null, result)
+  } catch (e) {
+    this.currentIndex = (this.currentIndex + 1) % this.urls.length
+    this.send(payload, callback, retries + 1)
+  }
+}
+
+module.exports = HttpListProvider
--- a/src/utils/HttpListProvider.js
+++ b/src/utils/HttpListProvider.js
@@ -10,11 +10,14 @@ function HttpListProvider(urls) {
 }
 
 HttpListProvider.prototype.send = async function(payload, callback, retries = 0) {
+  // save the currentIndex to avoid race condition
+  const currentIndex = this.currentIndex // eslint-disable-line prefer-destructuring
+
   if (retries === this.urls.length) {
     callback(new Error('Request failed for all urls'))
   }
 
-  const url = this.urls[this.currentIndex]
+  const url = this.urls[currentIndex]
 
   try {
     const result = await fetch(url, {
@@ -27,7 +30,7 @@ HttpListProvider.prototype.send = async function(payload, callback, retries = 0)
 
     callback(null, result)
   } catch (e) {
-    this.currentIndex = (this.currentIndex + 1) % this.urls.length
+    this.currentIndex = (currentIndex + 1) % this.urls.length
     this.send(payload, callback, retries + 1)
   }
 }
--- a/config/base.config.js
+++ b/config/base.config.js
@@ -6,6 +6,8 @@ const foreignNativeAbi = require('../abis/ForeignBridgeNativeToErc.abi')
 const homeErcAbi = require('../abis/HomeBridgeErcToErc.abi')
 const foreignErcAbi = require('../abis/ForeignBridgeErcToErc.abi')
 
+const rpcUrlsManager = require('../src/services/getRpcUrlsManager')
+
 const isErcToErc = process.env.BRIDGE_MODE && process.env.BRIDGE_MODE === 'ERC_TO_ERC'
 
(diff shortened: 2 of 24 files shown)


fvictorio:
Two more things:

First, there's something pending. Right now, if all URLs fail, the request (or web3 method) will fail. But the original issue says that it should wait and retry after a while. This is easy to add using `retry-promise`, but first let's check that everything is fine here, since this PR is hard enough to review as it is.

Second, I think the provider could be useful in other projects, so maybe it could be a separate project published in NPM and we could use it from there. Are you OK with that?

fvictorio:
> Currently I see that the functionality is too tied with 'home' and 'foreign' items. What exactly are you suggesting to move to a separate project?

Sorry, I was referring to the web3 provider. I think it's not tied to anything, so it should be usable in any project that could benefit from fallback URLs. I actually already moved it to its own repository, mainly for testing it, since doing that requires installing some very specific devDependencies. If you are OK with this idea, I can publish it in npm and update this PR to use it.

> Is my understanding correct and you in are trying to start every time with the first URL in the list even if was found that it does not response? Will it impact performance if we are trying to handle hundreds of transactions?
> Does it make sense to remember last successful URL in order to start from it rather than the first element of the URL list?

No, I think both the RpcUrlManager and the provider both use the last URL that worked. Let me know if you think that's not working.

fvictorio:
Sure, I just created the repo [here](https://github.com/fvictorio/http-list-provider) and is published in [npm](https://www.npmjs.com/package/http-list-provider). There are several things that could be done to improve it (support more node versions, support browser usage), but it can be used here as it is.

akolotov:
@rstormsf please take a look at this PR. Thanks!

src/utils/HttpListProvider.js
```diff
@@ -0,0 +1,38 @@
...
+  this.urls = urls
+  this.currentIndex = 0
+}
+
+HttpListProvider.prototype.send = async function(payload, callback, retries = 0) {
+  // save the currentIndex to avoid race condition
+  const currentIndex = this.currentIndex // eslint-disable-line prefer-destructuring
+
+  if (retries === this.urls.length) {
+    callback(new Error('Request failed for all urls'))
```
vbaranov: @fvictorio @akolotov what if for some reasons all of the RPC endpoints are not accessible in a short-term period, in this case, `send` method will fail here. What if to increase the level of service availability? I mean, to try to reach all RPC endpoints again. Introduce `retriesOuter` is equal, let's, say `3` - how many times `send` will try to reach all RPC endpoints. If all those tries will fail, then the method will fail. What do you think?

akolotov: Thanks @vbaranov for the comment! Actually it was requested in the original issue (https://github.com/poanetwork/bridge-nodejs/issues/28) and @fvictorio said (https://github.com/poanetwork/bridge-nodejs/pull/59#issuecomment-409891005) that is will added later after the review of main part of changes.

vbaranov: oh, I see. Nevertheless, the rest looks good to me

fvictorio:
@vbaranov Thanks for the comments. I added a commit to use the provider as a dependency. I also added a README to the repo.

Regarding the retry, I think we should create a separate issue and do it in another PR. The worst case scenario (all URLs fail) is not worse than what we have now (the single URL fails), so I think it's safe to merge this.

Issue #51, opened 2018-07-23
Stress Testing


patitonar:
Here are some ideas about how to measure the performance of the bridge. These tests could also be used to test the reliability of the bridge (if it continues to work under load or if it crashes).

1. Make a script that generates transactions in one side of the bridge and checks how many blocks are generated in the other side until the transaction is reflected. Run this script multiple times and compute statistics of the waiting time measured in blocks.
Pseudo-code:
```
beginBlockNumber = getBlockNumber(foreign)
sendTransaction(home)
waitForTransaction(foreign)
endBlockNumber = getBlockNumber(foreign)
numberOfBlocks = endBlockNumber - beginBlockNumber
```

2. Generate transactions and check the state of the rabbit queues (for example, average message waiting time).
3. Analyze logs to get the elapsed times of each step (from the time an event is processed by a watcher until all validators sign it, from there until the transaction is generated in the other side of the chain, etc.)

akolotov:
Thanks @patitonar for detailed description!
I have no questions for 2 and 3.
But for the item 1 we need to determine networks which are used for testing. E.g. if we use Ropsten, `numberOfBlocks` could be different from one run to another and it is OK since this test network will contain different number of concurrent transactions in different time periods.
That's why I would suggest to use two private chains for testing. Since they will not contain any other activities besides bridge transaction, `numberOfBlocks` could be approximately the same from one run to another. In this case we also should expect that the number of transactions included in every block is near to block gas limit. If so, the bridge completely fits in the network configuration.
What do you think?

fvictorio:
I think that if we use two private chains without activity, then the number of blocks won't be a useful measure. Maybe it could be useful for measuring time, but I'm not sure about that either.

I agree that if we do the tests using testnets that have other activities, the results won't be reproducible, but I think the numbers we obtain may be more realistic. And we could do several runs and average them to take into account the other "random" activity.

pablofullana:
On the one hand, I tend to think alike. I believe that measuring out in the wild will give us a more realistic sense of what end users could expect (and would experience).

On the other hand, there may be some situations is which we'd like to test in private networks. For example, when trying to replicate a known issue that may happen under certain circumstances (e.g. number of transactions). So, having this said, does it make any sense to make these target networks configurable?

akolotov:
thanks @fvictorio! from my point of view it more preferable to see how fast the bridge in handling events and sending transactions. E.g. the following test could be done for this with one bridge instance:
[...]

PR #60, opened 2018-08-10
Basic stress testing

fvictorio:
Related to #51.

This PR:

- Makes all scripts pipe their logs to disk. This make it easier to analyze them. Notice that this means that the logs to stdout are always prettified, but I don't think it's a problem since we'll have the raw log files (and if it is a problem, it's easy to fix).
- Adds a very basic script that analyzes these logs. This just computes some statistics on the time it passes between a signature request event is received and the corresponding transaction is sent, but there's a lot of room for improvement here (statistics in other events, the ability to select a time frame, we could even compute online statistics to use in a monitoring system, etc.).
- Adds a document that explains how to generate a block with N transactions. I did this manually, but I think it's something that may be automated if the need arises.

Doing this I found that the bridge fails when there are a lot of transactions in a block (it starts having issues around 200-300 txs). I think this happens because a lot of RPC requests are done to the node and, if one of them fails, the whole process fails. This is retried later, with the same or more transactions, and it fails again, and so on. I will open an issue about this so that we can discuss possible solutions.


package.json
```diff
@@ -5,15 +5,15 @@
   "main": "index.js",
   "scripts": {
     "lint": "eslint .",
-    "watcher:signature-request": "node src/watcher.js signature-request-watcher.config.js",
-    "watcher:collected-signatures": "node src/watcher.js collected-signatures-watcher.config.js",
-    "watcher:affirmation-request": "node src/watcher.js affirmation-request-watcher.config.js",
-    "sender:home": "node src/sender.js home-sender.config.js",
-    "sender:foreign": "node src/sender.js foreign-sender.config.js",
+    "watcher:signature-request": "node src/watcher.js signature-request-watcher.config.js | tee -a logs/watcher-signature-request.txt | pino",
```
akolotov: Is it possible to enable this only for a `dev` (or `testing`) environment? It is assuming that the bridge will be run by `systemd` facility on a production system this means that logs will be always redirected to `rsyslogd` and stored in the log files automatically.

package.json
```diff
@@ -5,15 +5,15 @@
...
+    "watcher:collected-signatures": "node src/watcher.js collected-signatures-watcher.config.js | tee -a logs/watcher-collected-signatures.txt | pino",
+    "watcher:affirmation-request": "node src/watcher.js affirmation-request-watcher.config.js | tee -a logs/watcher-affirmation-request.txt | pino",
+    "sender:home": "node src/sender.js home-sender.config.js | tee -a logs/sender-home.txt | pino",
+    "sender:foreign": "node src/sender.js foreign-sender.config.js | tee -a logs/sender-foreign.txt | pino",
     "dev": "concurrently -n 'watcher:signature-request,watcher:collected-signatures,watcher:affirmation-request,sender:home,sender:foreign' -c 'red,green,yellow,blue,magenta' 'npm run watcher:signature-request' 'npm run watcher:collected-signatures' 'npm run watcher:affirmation-request' 'npm run sender:home' 'npm run sender:foreign'",
     "test": "NODE_ENV=test mocha",
     "coverage": "NODE_ENV=test nyc --reporter=text --reporter=html mocha",
-    "postinstall": "npm install --prefix e2e"
+    "postinstall": "npm install --prefix e2e && mkdirp logs"
```
akolotov: Could you clarify why `mkdirp` is used instead of generic `mkdir` command?

akolotov:
> Adds a document that explains how to generate a block with N transactions. 
Great work! Thanks for documenting this!

> I found that the bridge fails when there are a lot of transactions in a block (it starts having issues around 200-300 txs)
Cool catch! This is exactly we need to run this tests!

package.json
fvictorio: No reason, really. `mkdirp` doesn't fail if the directory exists, but that can also be accomplished with `mkdir -p` (hence the module name). It's also more "portable", but I also used tee, so that's not a valid reason either. I will just use `mkdir` so that there are less dependencies.

package.json
fvictorio: Oh, yes. I will think a little about the best way to handle this. Some options are piping to an ad hoc module that prettifies (or doesn't) based on the `NODE_ENV` variable, or just removing the prettifier and making this a responsibility of the caller.

fvictorio:
> Did you measure how fast the bridge handled <200 transactions? I mean how many blocks did it take to keep all bridge responses?

I'm not sure I understand. If you are asking about the number of blocks, then I don't have a good measure since I used parity with instant seal as the foreign chain, so there's just one block per sent transaction.

akolotov:
> If you are asking about the number of blocks, then I don't have a good measure since I used parity with instant seal as the foreign chain, so there's just one block per sent transaction.

Ok. I see. Could you use Sokol as Foreign network for this test?

fvictorio:
I updated the `compute-stats` script to also analyze the times between a `CollectedSignatures` event is processed and the corresponding transaction is sent to the foreign chain.

I ran the bridge with a local chain as home and sokol as foreign, and created a block with 200 signature requests. The result is:

```
Signature Requests
{ mean: 3946.6, median: 3896.5, min: 3017, max: 5154 }
Collected Signatures
{ mean: 33059.9, median: 33495, min: 3086, max: 64177 }
```

This is just in time, not in blocks. To get some stats on blocks, we should log the number of the block in which a transaction was mined. We should also have some way of knowing the block number of the foreign chain at the time the `CollectedSignatures` event was seen in the home chain.

akolotov:
> Signature Requests
> { mean: 3946.6, median: 3896.5, min: 3017, max: 5154 }
> Collected Signatures
> { mean: 33059.9, median: 33495, min: 3086, max: 64177 }

Could you comment the results? Why do we have difference in 10 times?

fvictorio:
The main difference is that the second one corresponds to transactions made in sokol, while the first one is done in a local private chain.

But I'm not sure if this is enough. The transactions are only sent (not waited until they are mined). Maybe the node has some sort of rate limiting that kicks in here because 200 txs are being sent simultaneously?

akolotov:
Thansk @fvictorio. When you said "_200 txs are being sent simultaneously_" does it mean that they are really being sent in parallel threads? If so, could we measure difference in sending 25txs, 50txs, 100txs, 150 txs?

package.json
fvictorio: Please check the last changes to see if you are OK with that approach (I added a bash script that just checks the NODE_ENV variable and prettifies the output if the environment is not production).

fvictorio:
> When you said "200 txs are being sent simultaneously" does it mean that they are really being sent in parallel threads?

It means that they are all in the same block, so the bridge will process all of them together.

> If so, could we measure difference in sending 25txs, 50txs, 100txs, 150 txs?

Sure, I'll do that.

package.json
```diff
@@ -5,15 +5,15 @@
   "main": "index.js",
   "scripts": {
     "lint": "eslint .",
-    "watcher:signature-request": "node src/watcher.js signature-request-watcher.config.js",
-    "watcher:collected-signatures": "node src/watcher.js collected-signatures-watcher.config.js",
-    "watcher:affirmation-request": "node src/watcher.js affirmation-request-watcher.config.js",
-    "sender:home": "node src/sender.js home-sender.config.js",
-    "sender:foreign": "node src/sender.js foreign-sender.config.js",
+    "watcher:signature-request": "node src/watcher.js signature-request-watcher.config.js | tee -a logs/watcher-signature-request.txt | scripts/prettify.sh",
```
akolotov: If this is a generic command line, is it worth to introduce a script `start-bridge-worker.sh` with the following content?

```bash
#!/usr/bin/env bash

WORKERS_DIR="src/"
LOGS_DIR="logs/"

WORKER="${WORKERS_DIR}${1}"
CONFIG="${2}.config.js"
LOG="${LOGS_DIR}${2}.txt"

if [ "${NODE_ENV}" = "production" ]; then
  exec node ${WORKER} ${CONFIG}
else
  node ${WORKER} ${CONFIG} | tee -a ${LOG} | pino-pretty
fi
```

This script will manage processes on a production system more thriftily. 

So, `package.json` should be modified to contain:

```json
"watcher:signature-request": "scripts/start-bridge-worker.sh watcher.js signature-request-watcher",
"watcher:collected-signatures": "scripts/start-bridge-worker.sh watcher.js collected-signatures-watcher",
"watcher:affirmation-request": "scripts/start-bridge-worker.sh watcher.js affirmation-request-watcher",
"sender:home": "scripts/start-bridge-worker.sh sender.js home-sender.config.js",
"sender:foreign": "node scripts/start-bridge-worker.sh sender.js foreign-sender", 
```

The corresponding changes needs to be done in `scripts/compute-stats.js` to work with correct log file names.

fvictorio:
Here are some results.

First I used blocks with 25, 50, 100 and 200 transactions, using a local chain as foreign:

| txs | signature requests (mean time) | collected signatures (mean time) |
|-----|--------------------------------|----------------------------------|
| 25  | 1636.56                        | 526.16                           |
| 50  | 1106                           | 886.1                            |
| 100 | 2166.37                        | 1654.67                          |
| 200 | 3973.835                       | 3044.39                          |

And the same test, in sokol:

| txs | signature requests (mean time) | collected signatures (mean time) |
|-----|--------------------------------|----------------------------------|
| 25  | 614.48                         | 8755.76                          |
| 50  | 1116.26                        | 11094.76                         |
| 100 | 2217.23                        | 16913.72                         |
| 200 | 4001.635                       | 32307.34                         |

fvictorio:
It's worrying that the time grows linearly with the number of transactions: if the transactions are done in parallel, that shouldn't happen.

package.json
fvictorio: Totally agree with this. I added it (with some minor changes) and updated the `compute-stats` script.

My only concern is that the scripts will need to be executed always from the root of the project (since bash scripts use filenames relative to the cwd of the process). It's not a big issue, just something to keep in mind. (There are some [tricks](https://stackoverflow.com/questions/6659689/referring-to-a-file-relative-to-executing-script) that can be used, or the script can be ported to [shelljs](https://github.com/shelljs/shelljs), but I think this is enough for now).

src/events/processSignatureRequests.js
```diff
@@ -42,7 +42,10 @@ function processSignatureRequestsBuilder(config) {
           throw new Error(`RPC Connection Error: submitSignature Gas Estimate cannot be obtained.`)
         }
         logger.info(
-          { eventTransactionHash: signatureRequest.transactionHash },
+          {
+            eventTransactionHash: signatureRequest.transactionHash,
+            rpcResponse: e.message
```
akolotov: Please consider to add this changes as part of a separate PR. Since it is not related to stress testing but related to core functionality.

fvictorio: Sure. I rebased the branch to remove that commit.

rstormsf:
@fvictorio I thought we had a logic that if RPC call fails, it will be retried again

fvictorio:
Hi @rstormsf. That will be addressed in https://github.com/poanetwork/bridge-nodejs/issues/63 to avoid making this PR bigger.

akolotov:
If there are no more short plans to introduce additional changes to the repo as part of stress testing activity let's merge this PR

PR #69, opened 2018-08-18
Add a few helpful commands

rstormsf:
Update readme:
 - add commands
 - fix how to run tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ On `.env` file set `BRIDGE_MODE=ERC_TO_ERC`
   - `npm run sender:home`
   - `npm run sender:foreign`
 
-To send deposits to home contract run `node tests/sendUserTxToErcHome.js`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +84,8 @@ Command | Description
 --- | ---
 `rabbitmqctl list_queues` | List all queues
 `rabbitmqctl purge_queue home` | Remove all messages from `home` queue
+`rabbitmqctl status` | check if rabbitmq server is currently running  
+`rabbitmq-server`    | start rabbitMQ server
```
akolotov: Could you clarify if you met an issue and in order to solve it you needed to run `rabbitmq-server`? The same for the command `redis-server`.

Assuming that rabitmq and redis should be run in Docker. Could you specify in the description you introduced that `server will be run in Docker container`

akolotov: Ok. I see.

Issue #24, opened 2018-05-29
Enhance check for duplicated transactions handling


akolotov:
Depends on #37 

Watchers implemented:
- [x] signature-requests (#71)
- [x] collected-signatures
- [x] affirmation-request

Currently the bridge uses `estimateGas()` JSON-RPC API to check if the transaction was already processed. This check is not always accurate. For example, the exception will be raised by `estimateGas()` if the bridge instance operates with the private key which does not belong to the actual validator (e.g. the validator was removed from the contract), or if the message which is signed by the validator is corrupted etc. In such kind of cases the bridge will report that transaction was already processed but it is not true. It will slow down investigation of the issues since the log message does not reflect actual error.

That's why the check used `estimateGas()` could be extended with usage of methods provided by the bridge contract to provide more accurate logs:

1. The validator contract on the corresponding side could be inspected to make sure that the bridge is still is a validator by calling `isValidator()`:

https://github.com/poanetwork/poa-bridge-contracts/blob/93969cd934190649d42691f054676d0d16a9777e/contracts/upgradeable_contracts/U_BridgeValidators.sol#L71

2. `processDeposits` could be enhanced to check:
  * that the message was already handled by this validator by calling `messagesSigned` from Home Bridge contract
  * that the message was already relayed by calling `isAlreadyProcessed` from Home Bridge contract

3. `processCollectedSignatures` could be enhanced to check:
  * that enough number of signatures collected on the Home side by comparing the value from the `CollectedSignatures` event and value returned by `requiredSignatures` from the Validator contract on Foreign side. 
  * that the message was already relayed by calling `deposits` from Foreign Bridge contract

4. `processWithdraw` could be enhanced to check:
  * that the message was already handled by this validator by calling `withdrawalsSigned` from Home Bridge contract
  * that the message was already completely relayed by calling `isAlreadyProcessed` from Home Bridge contract

akolotov:
@pablofullana we need to decrease priority of this issue and do it after switching to v2.1 bridge contracts.

pablofullana:
@fvictorio @patitonar can suggest list below the set of changes needed to be done as per new contracts (`refactor_v1` branch)

akolotov:
Addressed in #71

PR #71, opened 2018-08-22
Enhance check for duplicated transactions handling

fvictorio:
This is an initial proposal on how to do #24. I added some custom errors so that the `processSignatureRequests` function knows why the gas estimation failed without having to parse the error messages.

Some things to notice here:

- I upgraded `http-list-provider`. The latest version throws a custom error so that we know if the gas estimation failed because of a connection error or a transaction error. You can see the change [here](https://github.com/fvictorio/http-list-provider/pull/1).
- I added two custom errors: `InvalidValidator` and `AlreadyProcessed`. Right now it throws `AlreadyProcessed` if the error was not caused by the provider and if it wasn't a problem with the validator. The goal is to make smarter checks.
- This is only done for `processSignatureRequests`, but something similar should be done in the other `process*` functions, avoiding as much duplication as possible. I only did it in one function to make sure that we all are OK this approach.

Let me know if you think this could work.


src/events/processSignatureRequests.js
```diff
@@ -1,11 +1,36 @@
...
+  try {
+    const gasEstimate = await method.estimateGas(options)
+    return gasEstimate
+  } catch (e) {
+    if (e instanceof HttpListProviderError) {
+      throw e
+    }
+
+    const validatorContractAddress = await bridgeContract.methods.validatorContract().call()
+    const validatorContract = new web3.eth.Contract(bridgeValidatorsABI, validatorContractAddress)
```
patitonar: I think it would be better to get the validator address and contract instance only once instead of every time a transaction gas estimate fails

patitonar: Besides previous comment, the approach looks good to me

rstormsf:
Merge conflicts

src/events/processSignatureRequests.js
```diff
@@ -1,11 +1,36 @@
...
+
+    const isValidator = await validatorContract.methods.isValidator(address).call()
+
+    if (!isValidator) {
+      throw new InvalidValidatorError(`${address} is not a validator`)
+    }
+
+    throw new AlreadyProcessedError(e.message)
```
akolotov: If we look at the contract code:
https://github.com/poanetwork/poa-bridge-contracts/blob/4e869c0c06dcff4d1d38ca65e9a62e973f5c3e97/contracts/upgradeable_contracts/BasicHomeBridge.sol#L43-L74
```
    function submitSignature(bytes signature, bytes message) external onlyValidator {
        // ensure that `signature` is really `message` signed by `msg.sender`
        require(Message.isMessageValid(message));
        require(msg.sender == Message.recoverAddressFromSignedMessage(signature, message));
        bytes32 hashMsg = keccak256(abi.encodePacked(message));
        bytes32 hashSender = keccak256(abi.encodePacked(msg.sender, hashMsg));

        uint256 signed = numMessagesSigned(hashMsg);
        require(!isAlreadyProcessed(signed));
        // the check above assumes that the case when the value could be overflew will not happen in the addition operation below
        signed = signed + 1;
        if (signed > 1) {
            // Duplicated signatures
            require(!messagesSigned(hashSender));
        } else {
            setMessages(hashMsg, message);
        }
        setMessagesSigned(hashSender, true);

        bytes32 signIdx = keccak256(abi.encodePacked(hashMsg, (signed-1)));
        setSignatures(signIdx, signature);

        setNumMessagesSigned(hashMsg, signed);

        emit SignedForUserRequest(msg.sender, hashMsg);

        uint256 reqSigs = requiredSignatures();
        if (signed >= reqSigs) {
            setNumMessagesSigned(hashMsg, markAsProcessed(signed));
            emit CollectedSignatures(msg.sender, hashMsg, reqSigs);
        }
    }
```
we can notice that `AlreadyProcessedError` could appear in two cases:
1. when this particular validator already sent the signature for this message (e.g. the database with block numbers was rolled back)
2. when enough number of signatures already collected (e.g. a new validator with empty database was set up).

It is worth to distinguish both these cases in the log and report them correctly in the logs.

In order to check that the exception is because the validator already signed this message it is necessary to
- calculate `keccak256(validator address, keccak256(message))`
- check that the call of `messagesSigned` with the data calculated above returns `true`

In order to check that the exception is because enough number of signatures already collected it is necessary to 
- calculate `keccak256(message)`
- call `numMessagesSigned` with the data calculated above
- check that the call of `isAlreadyProcessed` returns `true`

fvictorio: Yes, sorry, I didn't make myself clear. I only checked that the validator was valid as a minimal proof-of-concept of the implementation. The idea is to expand it with these other checks (and do it in the other functions).

I will continue with this approach then. (And thanks for the explanation on how to check the reason of the exception, that will help me a lot.)

src/events/processSignatureRequests.js
```diff
@@ -2,12 +2,65 @@ require('dotenv').config()
...
+      .estimateGas({
+        from: address
+      })
+    return gasEstimate
+  } catch (e) {
+    if (e instanceof HttpListProviderError) {
+      throw e
+    }
+
+    // Check if address is validator
```
fvictorio: @akolotov The order of this check is not the most efficient. They should be sorted from most likely to less likely. I think a better order would be:

1) Check if the tx was already processed by the other validators
2) Check if the tx was already signed by this validator
3) Check if the validator is still in the list of validators.

Do you agree?

src/events/processSignatureRequests.js
```diff
@@ -46,20 +99,40 @@ function processSignatureRequestsBuilder(config) {
...
+            logger.info(
+              { eventTransactionHash: signatureRequest.transactionHash },
+              `signatureRequest ${
+                signatureRequest.transactionHash
+              } was already processed by other validators`
+            )
+            return
+          } else {
+            logger.error(e, 'Unknown error while processing transaction')
+            throw e
```
fvictorio: The logic here is: if we `return`, then this transaction is ignored (this is done when the transaction was already signed or if it was processed by other validators). If we `throw`, the transaction will be retried later (this is done for connection errors, unknown errors and when the current address is not a validator).

I'm not sure about retrying when the address is not a validator, though. Maybe we should just `process.exit(1)` and be done with it?

src/events/processSignatureRequests.js
```diff
@@ -16,23 +16,18 @@ const { MAX_CONCURRENT_EVENTS } = require('../utils/constants')
...
       throw e
     }
 
     // Check if address is validator
-    const validatorContractAddress = await bridgeContract.methods.validatorContract().call()
-    const validatorContract = new web3.eth.Contract(bridgeValidatorsABI, validatorContractAddress)
-
     const isValidator = await validatorContract.methods.isValidator(address).call()
 
     if (!isValidator) {
```
akolotov: Yes, probably it is good idea to stop the process at all if the signer is not a validator. Could you use a specific exit code in order to we could distinguish different reasons why the process was stopped.
If it will be done in such manner it makes sense to forbid to run the watcher at all if the private key used to sign transactions does not belong to any address from the validators list.

fvictorio: > Could you use a specific exit code in order to we could distinguish different reasons why the process was stopped.

Sure.

> If it will be done in such manner it makes sense to forbid to run the watcher at all if the private key used to sign transactions does not belong to any address from the validators list.

You mean checking this during the initialization of the watcher? I totally agree with that.

akolotov: > You mean checking this during the initialization of the watcher? I totally agree with that.
Yes

fvictorio:
I moved the `estimateGas` to another module and added tests for it.

@akolotov I think we should merge this PR (if everything is OK) and make the proper changes for the other two watchers in separate PRs.

I added a checklist to #24 to keep track of this.

I also created #78 for the part of checking if the validator is valid.

akolotov:
I think we need to merge for all changes to address this functionality at once. Do you have any objections why we should split it and merge in different PRs? I do not see any other PRs which are in progress and require re-base to the code providing under this PR.

fvictorio:
> I think we need to merge for all changes to address this functionality at once. Do you have any objections why we should split it and merge in different PRs? I do not see any other PRs which are in progress and require re-base to the code providing under this PR.

No, it's OK. I'll push the rest of the work in this branch.

fvictorio:
@akolotov I added the extra checks for the `CollectedSignatures` watcher. I will work on the `AffirmationRequest` next, but you can review the new changes if you want.

Two comments:
- I had to add a `parseMessage` function to extract the transaction hash that is used to check if the event was already processed. The problem with this is that it duplicates the same logic in the `Message.sol` library, but we can't use that library function from the code (unless it's exposed by the contract like we did for `requiredMessageLength` in `BasicHomeBridge.sol`). Just duplicating the logic was the easiest alternative, but it's far less maintainable.
- So far I've added new errors for the different scenarios. But I think many of these errors could be grouped in one, with different codes. Something like `IncompatibleContractsError` with a code for an invalid validator, another for the required number of signatures, etc. (cc @patitonar).

fvictorio:
I added the extra checks for `processAffirmationRequests`.

fvictorio:
Added the checks for `processTransfers`.

src/events/processCollectedSignatures/estimateGas.js
```diff
@@ -0,0 +1,51 @@
...
+    const gasEstimate = await foreignBridge.methods
+      .executeSignatures(v, r, s, message)
+      .estimateGas()
+    return gasEstimate
+  } catch (e) {
+    if (e instanceof HttpListProviderError) {
+      throw e
+    }
+
+    const requiredSignatures = await foreignBridge.methods.requiredSignatures().call()
```
akolotov: I don't really understand how it works. `requiredSignatures` is a method from Validators contract whereas `foreignBridge` contract is used here.

src/events/processCollectedSignatures/index.js
```diff
@@ -52,20 +70,37 @@ function processCollectedSignaturesBuilder(config) {
...
+              )
+            } else if (e instanceof AlreadyProcessedError) {
+              logger.info(
+                { eventTransactionHash: colSignature.transactionHash },
+                `Already processed CollectedSignatures ${colSignature.transactionHash}`
               )
+              return
+            } else if (e instanceof IncompatibleContractError) {
+              logger.fatal(`The contract is not compatible: ${e.message}`)
+              process.exit(10)
```
akolotov: So far I see only one reason when `IncompatibleContractError` appears - when number of signatures collected on `Home` side less than number of signatures expected by the contract on `Foreign` side. It could be valid scenario when number of required signatures was changed and the bridge instance is trying to re-handle some messages (e.g.  database of handled blocks was rolled back). The bridge must not crash here, it is incorrect behavior.

src/events/processCollectedSignatures/estimateGas.js
```diff
@@ -0,0 +1,51 @@
...
+    if (requiredSignatures.toString() !== numberOfCollectedSignatures.toString()) {
```
akolotov: Number of collected signatures could be more than required. It is OK.
Here is the corresponding check from contracts:
```solidity
        uint256 requiredSignatures = _validatorContract.requiredSignatures();
        require(_vs.length >= requiredSignatures);
```
And I'd move this check after the check that the message already relayed because the second one is more probable.

src/events/processCollectedSignatures/estimateGas.js
```diff
@@ -0,0 +1,51 @@
...
+      throw new IncompatibleContractError('The number of collected signatures does not match')
+    }
+
```
akolotov: If you look at the contract code you will see that every signature in the message is checked to be sure that:
* the corresponding signature was made by a validator;
* this signature is mentioned only once in the array;

```solidity
        for (uint256 i = 0; i < requiredSignatures; i++) {
            address recoveredAddress = ecrecover(hash, _vs[i], _rs[i], _ss[i]);
            require(_validatorContract.isValidator(recoveredAddress));
            if (addressArrayContains(encounteredAddresses, recoveredAddress)) {
                revert();
            }
            encounteredAddresses[i] = recoveredAddress;
        }
```
So let's add this check as well. The check that we will get a singature which belongs to an account and it is not validator any more is still probable. You could even consider to add this check just after the line:

```javascript
const signature = await homeBridge.methods.signature(messageHash, index).call()
```

src/events/processCollectedSignatures/estimateGas.js
```diff
@@ -0,0 +1,51 @@
...
+    const { txHash } = parseMessage(message)
+    const alreadyProcessed = await foreignBridge.methods.relayedMessages(txHash).call()
+    if (alreadyProcessed) {
+      throw new AlreadyProcessedError()
+    }
+
+    // Check if address is validator
+    const isValidator = await validatorContract.methods.isValidator(address).call()
```
akolotov: This check is not needed. The method `executeSignatures` could be called by any account so will not fail if the code is executed by an account which is not in the list of validators.

src/events/processCollectedSignatures/index.js
fvictorio: What would be the correct behavior here? If we ignore the message, it will never be processed. If we just throw, it will keep retrying forever.

akolotov: For scenario when number of signatures collected on `Home` side less than number of signatures expected by the contract on `Foreign` side we need report about this in the log and handle it by the same as we handle e.g. `AlreadySigned`.

src/events/processCollectedSignatures/estimateGas.js
fvictorio: In this case, maybe is better to just ask the contract. The problem is that this logic is in the `Message` library and we can't call it directly. The same thing happened to me with the `parseMessage` logic.

Since it's the third time that this happens, would it make sense to expose all the library functions somehow? I see two ways in which this could be done:

1. Make all methods `internal` and add methods with the same name in the bridge contracts that just call the library.
2. Deploy and link the library and drop the `internal`s.

**2** looks better to me, but it would mean modifying the deploy script.

akolotov: I would avoid big amount of changes on the contracts side since we already had a security audit for them. 

If we just ask the contract we still get no point why we fails since two check are combined there:
```solidity
require(_validatorContract.isValidator(recoveredAddress));
```
 and
```solidity
if (addressArrayContains(encounteredAddresses, recoveredAddress)) {
    revert();
}
```

My suggestion is the following:
* Since we generate the list of signatures before passing it to the contract, it is not necessary to check for signature duplication in `processCollectedSignatures/estimateGas.js`.
* But we need to make sure that the signature in the list is for an actual validator. That's why we need to restore every validator address from the signatures we received and call `isValidator` for them.

akolotov: Yes, at least in the nearest future (release of `native-to-erc`, `erc-to-erc` and `erc-to-native`) we will not change contracts a lot.

src/events/processCollectedSignatures/index.js
akolotov: > So, in this case, we log the occurrence and don't process the message.

Correct.

>  I guess it's also something that could be monitored.

Yes, we should see the corresponding record at least in the logs.

src/events/processCollectedSignatures/estimateGas.js
fvictorio: This was a mistake. I must've missed this when I tested this manually. Thanks, fixed.

src/events/processCollectedSignatures/index.js
fvictorio: Done.

src/events/processCollectedSignatures/estimateGas.js
fvictorio: Done.

src/events/processCollectedSignatures/estimateGas.js
fvictorio: I passed the signatures to the `estimateGas` function to keep the structure of error checking. There it recovers the address for every signature and checks that is from a validator.

src/events/processCollectedSignatures/estimateGas.js
fvictorio: Done, thanks.

fvictorio:
@akolotov Thanks for the comments, I applied the checks.

src/events/processCollectedSignatures/index.js
```diff
@@ -92,11 +90,8 @@ function processCollectedSignaturesBuilder(config) {
               )
               return
             } else if (e instanceof IncompatibleContractError) {
-              logger.fatal(`The contract is not compatible: ${e.message}`)
-              process.exit(10)
-            } else if (e instanceof InvalidValidatorError) {
-              logger.fatal({ address: VALIDATOR_ADDRESS }, 'Invalid validator')
-              process.exit(10)
+              logger.error(`The contract is not compatible: ${e.message}`)
```
akolotov: It is not correct to say that "the contract is not compatible". The issue actually is with data collected by the contract.

src/events/processCollectedSignatures/estimateGas.js
```diff
@@ -26,22 +26,27 @@ async function estimateGas({
...
+
+    // check if all the signatures were made by validators
+    for (const signature of signatures) {
+      const address = web3.eth.accounts.recover(message, signature)
+      const isValidator = await validatorContract.methods.isValidator(address).call()
 
-    if (!isValidator) {
-      throw new InvalidValidatorError(`${address} is not a validator`)
+      if (!isValidator) {
+        throw new IncompatibleContractError(`Message signed by ${address} that is not a validator`)
```
akolotov: Seems that we can use `InvalidValidatorError` here.

src/events/processCollectedSignatures/index.js
```diff
@@ -73,12 +71,12 @@ function processCollectedSignaturesBuilder(config) {
             gasEstimate = await estimateGas({
               foreignBridge,
               validatorContract,
+              signatures,
```
akolotov: Why do you pass `signatures` to the function if you already pass `v`, `r` and `s` components of these signatures?

akolotov:
Could you run stress-testing as soon as all changes made under this PR are approved? Thanks!

src/events/processCollectedSignatures/index.js
fvictorio: When I tried it in the repl, I couldn't make `eth.accounts.recover` work with `v`, `r` and `s`. But I will give it another try.

fvictorio: Done.

src/events/processCollectedSignatures/index.js
fvictorio: I grouped both problems, the number of signatures being less than required and the signatures being incorrect. In both cases, the message is ignored and the situations is logged.

fvictorio:
> Could you run stress-testing as soon as all changes made under this PR are approved? Thanks!

I applied the suggested changes. I will run the stress-testing now.

fvictorio:
Stress testing results:

```
Signature Requests
{
  "count": 1000,
  "mean": 12238.425,
  "median": 12478,
  "min": 5911,
  "max": 17914
}
Collected Signatures
{
  "count": 1000,
  "mean": 8095.244,
  "median": 8102,
  "min": 7829,
  "max": 8323
}
```

akolotov:
> Stress testing results:

What values are for the version without your changes?

src/events/processCollectedSignatures/estimateGas.js
```diff
@@ -44,8 +43,8 @@ async function estimateGas({
     }
 
     // check if all the signatures were made by validators
-    for (const signature of signatures) {
-      const address = web3.eth.accounts.recover(message, signature)
+    for (let i = 0; i < v.length; i++) {
+      const address = web3.eth.accounts.recover(message, web3.utils.toHex(v[i]), r[i], s[i])
```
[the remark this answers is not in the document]
fvictorio: Yes, it seems to work:

```
> signed = eth.accounts.sign('hello', '0x8e829f695aed89a154550f30262f1529582cc49dc30eff74a6b491359e0230f9')
{ message: 'hello',
  messageHash: '0x50b2c43fd39106bafbba0da34fc430e1f91e3c96ea2acee2bc34119f92b37750',
  v: '0x1b',
  r: '0x92fa2d2fda7f364f6905fcc2a05d8aad831b24d1526be02b32ffb7ff43346e5c',
  s: '0x09f58dd6d96add2c6f294ab60aa84f138ecd877fcef075df4b3014e28a0236e0',
  signature: '0x92fa2d2fda7f364f6905fcc2a05d8aad831b24d1526be02b32ffb7ff43346e5c09f58dd6d96add2c6f294ab60aa84f138ecd877fcef075df4b3014e28a0236e01b' }
> eth.accounts.recover(signed.messageHash, signed.v, signed.r, '0x9f58dd6d96add2c6f294ab60aa84f138ecd877fcef075df4b3014e28a0236e0', true)
'0xaaB52d66283F7A1D5978bcFcB55721ACB467384b'
```

Notice that I removed the first 0 of the `s` value.

fvictorio:
These are the times without the changes:

```
Signature Requests
{
  "count": 1000,
  "mean": 11033.719,
  "median": 11071.5,
  "min": 5401,
  "max": 16357
}
Collected Signatures
{
  "count": 1000,
  "mean": 6808.935,
  "median": 7202.5,
  "min": 846,
  "max": 7619
}
```

src/events/processCollectedSignatures/estimateGas.js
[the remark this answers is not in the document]
akolotov: Great! Thanks!

akolotov:
Collected Signatures -> "min" increased almost in 10 times. Do you have ideas why?

fvictorio:
There seems to be quite a variance with the `min` value. I ran the test once more, with and without changes, and the min value was shorter with the changes and longer without the changes:

|statistic|without changes|with changes|
|----------|------------------|---------------------|
|count|1000|1000|
|mean|7037.111|6318.651|
|median|6951|7036|
|min|4443|1506|
|max|8011|7927|

I think the min value is highly dependent on the order in which transactions are processed, so it may not be a very good indicator.

akolotov:
Ok. In general this changes do not introduce significant degradation. So I approve the changes to be merged. Thanks for your work!

Issue #64, opened 2018-08-15
Processing fails when a block has hundreds of transactions


fvictorio:
See discussion in https://github.com/poanetwork/bridge-nodejs/pull/60.

Create a block with 300 transactions. After processing some of them, the RPC responds with an error. This causes all transactions to be retried later, and fail again.

Possible solutions:

- Retry when a request fails (related to https://github.com/poanetwork/bridge-nodejs/issues/63).
- Set a max size to the number of events that are processed at the same time (this may need some architectural changes).

rstormsf:
Here is what I think is happening:
1. Worker has a loop to ask every X seconds( 5 seconds) to query for new blocks
2. Since our  previous block wasn't fully processed, we still have local db pointed to old block
3. The new RPC request(timeout) should be less than next run to query for new blocks.
So if there are failed RPC requests, all workers should not probably try to fetch information about new transactions.

akolotov:
Addressed in #72.

Issue #63, opened 2018-08-13
Retry request when all RPC URLs fail


fvictorio:
Related to #28.

If all the RPC URLs fail, retry after some time.

We can use [`promise-retry`](https://www.npmjs.com/package/promise-retry) for this. Also, this should be done both for the raw HTTP requests and in the web3 provider.

PR #72, opened 2018-08-23
Limit number of concurrent events that are processed

fvictorio:
This should fix, or at least improve, the problem described in #64.

The cause of the problem was that making a lot of concurrent `estimateGas` calls were causing the node to fail (I confirmed this by making a simple contract with a loop and calling it hundreds of times in parallel; let me know if you'd like to see the code). When one of this transactions failed, the whole list of events was discarded. When later the same (or more) events were received, the problem appeared again, and so on.

The fix is to limit the number of concurrent events that are processed (right now to 50). With this fix, we could process a block with 8000 txs without failures.

Doing #63 might have fixed this too, but even with that I think it's a good idea to avoid choking the RPC endpoint. Besides, I'm not 100% sure that #63 would've been enough (if all the retries have the same sucessive delays, each the RPC would be choked again, albeit with less transactions each time).

Some time measurements:

* **Signature requests mean time**
  * 2000 txs: 25383 ms
  * 4000 txs: 49654 ms
  * 8000 txs: 93609 ms
* **Collected signatures mean time**
  * 2000 txs: 12939 ms
  * 4000 txs: 22013 ms
  * 8000 txs: 47697 ms


fvictorio:
Maybe I should also measure the difference between the processing time with and without this change (using a number of transactions that can be handled by the bridge prior to this fix). I'll do that and post the results here.

rstormsf:
ci is failing

src/utils/constants.js
```diff
@@ -1,3 +1,4 @@
 module.exports = {
-  EXTRA_GAS_PERCENTAGE: 0.25
+  EXTRA_GAS_PERCENTAGE: 0.25,
+  MAX_CONCURRENT_EVENTS: 50
```
patitonar: Perhaps we should run some performance tests to find which is the most efficient value for this. Maybe we can address this in a separate issue.

PR #73, opened 2018-08-28
Retry request when all RPC URLs fail

### LINKED ISSUE #63: Retry request when all RPC URLs fail

fvictorio:
Closes #63. Do not merge until https://github.com/fvictorio/http-list-provider/pull/2 is merged.

Use [promise-retry](https://github.com/IndigoUnited/node-promise-retry) to retry the RPC calls when all URLs fail.

Right now this uses the default retry setup: 10 retries with exponential backoff. I don't think this is the best configuration: it will wait `2^10` seconds before failing for sure, which is a lot. Let me know what you think is the best setup; check `promise-retry` documentation to see the possible options.

This also moves all the web3 instantiations to one module. I did this to avoid duplicating the retry configuration in several places.


src/tx/sendTx.js
```diff
@@ -51,6 +51,12 @@ async function sendRawTx({ chain, params, method }) {
         id: Math.floor(Math.random() * 100) + 1
       })
     })
+
+    if (!response.ok) {
```
akolotov: Could you share more details why this is needed now? Why did it work before?

fvictorio: The RPC endpoint always returns a 200 status code, even if the transaction failed (the error is inside the response body). But this could fail for connection reasons: for example, if the RPC is behind a proxy and the proxy returns a 404 for some reason.

I think this worked before because we used local or reliable endpoints, but when I checked with failing URLs the lack of this check was causing issues.

fvictorio:
> is it possible to configure factor as 1.4? If yes, lets have
> 
> retries: 20
> factor: 1.4
> maxTimeout: 360000
> randomize: true
> is it OK?

Sure!

src/tx/sendTx.js
akolotov: thanks for explanation. I am OK with this.

fvictorio:
> maxTimeout: 360000

Just in case, that number is 6 minutes. It's that what you meant or do you want it to be 1 hour?

akolotov:
6 minutes timeout is big enough.

PR #77, opened 2018-09-04
Add LICENSE

### LINKED ISSUE #76: License missing

fvictorio:
Closes #76.

@akolotov I used MIT because is the same that is used in the Token Wizard. Let me know if a different one should be used.


igorbarinov:
For products, we use GPL3 
For libraries, we use MIT/Apache

Let's have GPL3 for the bridge

fvictorio:
@igorbarinov Done.

PR #85, opened 2018-09-26
Add support ERC20 to Native

### LINKED ISSUE #79: Update README
Please re-check and update README.
Things I noticed:

1. https://github.com/poanetwork/bridge-nodejs#how-to-use - "Deploy bridge contracts" says to use `v2` branch of bridge contracts, while it should be `refactor_v1`. (deployment steps are not working for `v2` anyway)
2. https://github.com/poanetwork/bridge-nodejs#run-the-processes - `sendUserTx*` scripts are in `scripts/`, not in `tests/`. They also require a copy of `.env` to be in `scripts/` to run properly.

### LINKED ISSUE #81: Support of new bridge mode: ERC20 to Native
In order to implement a new bridge mode described in https://hackmd.io/s/Bk9icHBd7, https://github.com/poanetwork/poa-bridge-contracts/issues/79 was created for the changes on the bridge contracts side. 

This issue was created to address necessary modification on bridge process side. The intermediate list of changes: 
- [ ] Adapt the initialization code to determine the bridge mode. The current version of code uses `BRIDGE_MODE` parameter but it makes sense to consider selectors of contracts' methods specific for particular bridge mode   
- [ ] Improve configs
- [ ] improve scripts
- [ ] Add e2e tests
- [ ] Update README

patitonar:
This PR includes the following changes:
- Add ERC20-to-Native abis
- Add workers support for ERC20-to-Native mode
- Update related scripts to support new mode
- Update README, including fixes proposed on #79 
- Add e2e test for ERC20-to-Native

Once https://github.com/poanetwork/poa-bridge-contracts/pull/82 is merged, we should update contract submodule and `e2e/Dockerfile` to point to correct contract branch.

Closes #81 , closes #79


fvictorio:
Just tested it and everything seemed to work fine.

akolotov (changes requested):
The gas priced received from the oracle is in GWei whereas the gas price from the contract is in Wei:

https://github.com/poanetwork/bridge-nodejs/blob/8359556d9e9976a9388720d97de70e1bd7b35af8/src/services/gasPrice.js#L43-L57

It will cause incorrect conversion in the following code: 

https://github.com/poanetwork/bridge-nodejs/blob/8359556d9e9976a9388720d97de70e1bd7b35af8/src/tx/sendTx.js#L18-L29

It make sense to have the fallback gas price in the bridge config set in Wei as well.

Therefore consider to modify `fetchGasPriceFromOracle` to return the value converted from GWei to Wei.

patitonar:
Thanks @akolotov , good catch! I updated `sendTx` to remove conversion from Gwei, updated `fetchGasPriceFromOracle` to return value converted to Wei, updated scripts to pass gasPrice in Wei and set fallback env variables in Wei. Please check last commit

akolotov:
Since you have merged new README.md file please update it with new meaning of configuration parameters (e.g. HOME_GAS_PRICE_FALLBACK and FOREIGN_GAS_PRICE_FALLBACK are now in wei).

scripts/erc20_to_erc20/sendHome.js
```diff
@@ -80,7 +80,7 @@ async function main() {
         privateKey: USER_ADDRESS_PRIVATE_KEY,
         data,
         nonce,
-        gasPrice: '1',
+        gasPrice: '1000000000',
```
akolotov: Can we introduce a parameter to configure the gas price which will be used for testing transactions?

scripts/erc20_to_native/sendForeign.js
```diff
@@ -0,0 +1,95 @@
...
+const Web3Utils = require('web3-utils')
+const rpcUrlsManager = require('../../src/services/getRpcUrlsManager')
+const { sendTx, sendRawTx } = require('../../src/tx/sendTx')
+
+const {
+  USER_ADDRESS,
+  USER_ADDRESS_PRIVATE_KEY,
+  FOREIGN_BRIDGE_ADDRESS,
+  FOREIGN_MIN_AMOUNT_PER_TX,
+  ERC20_TOKEN_ADDRESS
```
akolotov: Can `ERC20_TOKEN_ADDRESS` will be extracted from the bridge contract?

scripts/erc20_to_native/sendHome.js
```diff
@@ -35,9 +35,9 @@ async function main() {
         privateKey: USER_ADDRESS_PRIVATE_KEY,
         data: '0x',
         nonce,
-        gasPrice: '1',
+        gasPrice: '1000000000',
         amount: HOME_MIN_AMOUNT_PER_TX,
```
akolotov: Could we make sure that `HOME_MIN_AMOUNT_PER_TX` is greater than the "Minimum amount per transaction" limit before starting to send transactions?

.env.example
```diff
@@ -13,12 +13,12 @@ VALIDATOR_ADDRESS_PRIVATE_KEY=
 
 HOME_GAS_PRICE_ORACLE_URL=https://gasprice.poa.network/
```
akolotov: Here is one of the comment we received during the security audit is: https://hackmd.io/s/Sk6P7Uqtm#2-Bad-private-data-storing. 
It is related to an environment when watcher and senders are run in the docker containers.
In order to address the comment I suggest the following:
1. For the environment when the watchers and senders are run on the same system (not containerized) `VALIDATOR_ADDRESS` is redundant - it can be calculated from the `VALIDATOR_ADDRESS_PRIVATE_KEY`. The official recommendation will be to put have `VALIDATOR_ADDRESS_PRIVATE_KEY` only in the configuration file. Moreover it will reduce the number of issues when a developer puts the validator address which does not correspond to the private key.
2. If the watcher finds that `VALIDATOR_ADDRESS_PRIVATE_KEY` is not configured (it means that it run in the container), it should use `VALIDATOR_ADDRESS`.

As soon as these changes are done, we will re-write the deployment playbooks as so they will store `VALIDATOR_ADDRESS_PRIVATE_KEY` in sender's containers only. Also the bridge startup script will calculate `VALIDATOR_ADDRESS` from `VALIDATOR_ADDRESS_PRIVATE_KEY` programmatically and initialize the watcher's contaners as so they will have `VALIDATOR_ADDRESS` without mentioning `VALIDATOR_ADDRESS_PRIVATE_KEY`.

What do you think?

scripts/native_to_erc20/sendForeign.js
```diff
@@ -1,10 +1,10 @@
...
+  path: path.join(__dirname, '../../.env')
 })
 const Web3Utils = require('web3-utils')
-const { web3Foreign } = require('../src/services/web3')
-const { sendTx, sendRawTx } = require('../src/tx/sendTx')
+const { web3Foreign } = require('../../src/services/web3')
+const { sendTx, sendRawTx } = require('../../src/tx/sendTx')
 
 const {
   USER_ADDRESS,
```
akolotov: Can ERC20_TOKEN_ADDRESS will be extracted from the bridge contract?

scripts/erc20_to_erc20/sendForeign.js
```diff
@@ -1,11 +1,11 @@
...
 })
 const Web3 = require('web3')
 const Web3Utils = require('web3-utils')
-const rpcUrlsManager = require('../src/services/getRpcUrlsManager')
-const { sendTx, sendRawTx } = require('../src/tx/sendTx')
+const rpcUrlsManager = require('../../src/services/getRpcUrlsManager')
+const { sendTx, sendRawTx } = require('../../src/tx/sendTx')
 
 const {
   USER_ADDRESS,
```
akolotov: Can ERC20_TOKEN_ADDRESS will be extracted from the bridge contract?

scripts/erc20_to_erc20/sendHome.js
```diff
@@ -1,11 +1,11 @@
...
 })
 const Web3 = require('web3')
 const Web3Utils = require('web3-utils')
-const rpcUrlsManager = require('../src/services/getRpcUrlsManager')
-const { sendTx, sendRawTx } = require('../src/tx/sendTx')
+const rpcUrlsManager = require('../../src/services/getRpcUrlsManager')
+const { sendTx, sendRawTx } = require('../../src/tx/sendTx')
 
 const {
   USER_ADDRESS,
```
akolotov: Can BRIDGEABLE_TOKEN_ADDRESS will be extracted from the bridge contract?

.env.example
fvictorio: Let me see if I understood this correctly:

- Senders extract the `VALIDATOR_ADDRESS_PRIVATE_KEY` from the environment variables and use it to calculate the validator address.
- Watchers extract `VALIDATOR_ADDRESS` from the environment: this will be injected by the startup script. Also watchers should work fine if the private key is not present in the environment variables (since they don't use it anyway)

Did I get this right?

akolotov: Right.
And watchers should work in the case if `VALIDATOR_ADDRESS` is not specified (the case when watchers and senders are run in the same environment - not containerized) - `VALIDATOR_ADDRESS_PRIVATE_KEY` is used by watchers to calculate the validator address.

fvictorio: I see, thanks!

scripts/erc20_to_native/sendForeign.js
fvictorio: Done.

scripts/native_to_erc20/sendForeign.js
fvictorio: Done.

scripts/erc20_to_erc20/sendForeign.js
fvictorio: Done.

scripts/erc20_to_erc20/sendHome.js
fvictorio: Done.

scripts/erc20_to_erc20/sendHome.js
patitonar: Done. Added `HOME_TEST_TX_GAS_PRICE` and `FOREIGN_TEST_TX_GAS_PRICE`

scripts/erc20_to_native/sendHome.js
patitonar: Done. Added that check on every `sendHome/Foreign` script

scripts/erc20_to_native/sendForeign.js
```diff
@@ -30,6 +31,8 @@ async function main() {
   const poa20 = new web3Foreign.eth.Contract(ERC20_ABI, ERC20_TOKEN_ADDRESS)
 
   try {
+    await isValidAmount(FOREIGN_MIN_AMOUNT_PER_TX, bridge)
```
akolotov: the minimum amount per transaction is not applicable for the Foreign side of the bridge in `erc20-to-native` mode.

scripts/erc20_to_erc20/sendForeign.js
```diff
@@ -30,6 +31,8 @@ async function main() {
   const poa20 = new web3Foreign.eth.Contract(ERC20_ABI, ERC20_TOKEN_ADDRESS)
 
   try {
+    await isValidAmount(FOREIGN_MIN_AMOUNT_PER_TX, bridge)
```
akolotov: the minimum amount per transaction is not applicable for the Foreign side of the bridge in `erc20-to-erc20` mode.

akolotov (changes requested):

.env.example
akolotov: @fvictorio when could we expect these set of changes? I need  to send them for review to the security auditor to response on the opened finding.

fvictorio: Done in 6575f0fd2cc8883f042bf2cd94d63840ad8031f0, but notice that the signature requests watcher still needs the private key:

https://github.com/poanetwork/token-bridge/blob/2fad9b9cca14b3eedbcecc8c931c845906c70bc4/src/events/processSignatureRequests/index.js#L60

Do the watchers **have** to work without the private key? Maybe there's some way to rewrite this so we don't need it.

scripts/erc20_to_native/sendForeign.js
patitonar: Fixed

scripts/erc20_to_erc20/sendForeign.js
patitonar: Fixed

.env.example
akolotov: Ok. Let's leave the ability to work with the private key for the `signature-requests` watcher. Later we will think how to split this logic.

akolotov:
As part of the work to address issues raised during the security audit, please fix https://github.com/poanetwork/token-bridge/issues/107 in this PR

scripts/privateKeyToAddress.js
```diff
@@ -0,0 +1,12 @@
+const { privateKeyToAddress } = require('../src/utils/utils')
+
+const privateKey = process.argv[2]
```
akolotov: the private key should be get from the environment variable `VALIDATOR_ADDRESS_PRIVATE_KEY` as it was described in https://github.com/poanetwork/token-bridge/issues/107.

fvictorio: Done.

src/sender.js
```diff
@@ -142,13 +150,18 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
     logger.info(`Finished processing msg`)
```
akolotov: I think the level should `debug` here. Since it is not required to report that the message was processed successfully if no errors appear.

src/sender.js
```diff
@@ -131,7 +136,10 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
       }
     })
 
+    logger.debug('Updating nonce')
     await updateNonce(nonce)
+
+    logger.debug('Releasing lock')
     await lock.unlock()
 
     if (failedTx.length) {
```
akolotov: should it be `warn` or `error`?

src/sender.js
```diff
@@ -76,17 +76,22 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
...
     let minimumBalance = null
     const failedTx = []
 
+    logger.debug(`Sending ${txArray.length} transactions`)
     await syncForEach(txArray, async job => {
       const gasLimit = addExtraGas(job.gasEstimate, EXTRA_GAS_PERCENTAGE)
 
       try {
+        logger.debug(`Sendint transaction with nonce ${nonce}`)
         const txHash = await sendTx({
```
akolotov: I think it is reasonable to have a logging for all transaction' raw data (as well as whole JSON sent to a RPC node). Consider to have an environment variable flag (e.g. `RPC_TRACES=true/false`) that will enable or disable this extended logging.

src/sender.js
```diff
@@ -142,13 +150,18 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
     logger.info(`Finished processing msg`)
 
     if (insufficientFunds) {
+      logger.warn(
+        'Insufficient funds. Stop sendint transactions until the account has the minimum balance'
+      )
       channel.close()
       waitForFunds(web3Instance, VALIDATOR_ADDRESS, minimumBalance, resume)
```
akolotov: since this method requests data from the chain it make sense to add logging there in order to see if it stuck or not.

src/events/processTransfers/index.js
```diff
@@ -29,10 +29,15 @@ function processTransfersBuilder(config) {
       validatorContract = new web3Home.eth.Contract(bridgeValidatorsABI, validatorContractAddress)
```
akolotov: Please log the received value of `validatorContractAddress`.

src/events/processTransfers/index.js
```diff
@@ -29,10 +29,15 @@ function processTransfersBuilder(config) {
...
     }
 
+    rootLogger.debug(`Processing ${transfers.length} Transfer events`)
     const callbacks = transfers.map(transfer =>
       limit(async () => {
         const { from, value } = transfer.returnValues
 
+        const logger = rootLogger.child({
```
akolotov: Is my understanding correct and we don't log any information about the transaction that raised this event below (similarly we have in `processAffirmationRequests`)? If it is true, let's add here a log message for this.

src/events/processTransfers/index.js
```diff
@@ -29,10 +29,15 @@ function processTransfersBuilder(config) {
...
+          eventTransactionHash: transfer.transactionHash
+        })
+
         let gasEstimate
         try {
           gasEstimate = await estimateGas({
```
akolotov: It makes sense to add logging within `estimateGas` to return the value received by the the corresponding JSON-RPC call. It will also allow to check that it was not stuck there.
Moreover there are lot of call to RPC in `estimateGas`. So it seems reasonable to add logging after every call there.

src/events/processSignatureRequests/index.js
```diff
@@ -35,12 +35,17 @@ function processSignatureRequestsBuilder(config) {
...
 
+        const logger = rootLogger.child({
+          eventTransactionHash: signatureRequest.transactionHash
+        })
+
         logger.info(
-          { eventTransactionHash: signatureRequest.transactionHash, sender: recipient, value },
+          { sender: recipient, value },
           `Processing signatureRequest ${signatureRequest.transactionHash}`
         )
```
akolotov: consider to add a log after `sign` in the code below.

src/events/processCollectedSignatures/index.js
```diff
@@ -46,11 +47,12 @@ function processCollectedSignaturesBuilder(config) {
...
+          eventTransactionHash: colSignature.transactionHash
+        })
+
         if (authorityResponsibleForRelay === web3Home.utils.toChecksumAddress(VALIDATOR_ADDRESS)) {
-          logger.info(
-            { eventTransactionHash: colSignature.transactionHash },
-            `Processing CollectedSignatures ${colSignature.transactionHash}`
-          )
+          logger.info(`Processing CollectedSignatures ${colSignature.transactionHash}`)
           const message = await homeBridge.methods.message(messageHash).call()
```
akolotov: consider to add logging after this to check that we don't stuck here. The same is for the method to get signatures below.

akolotov (changes requested):
Since we work on the improving the logging anyway, I added requests to extend the logging for better identification of possible breakdowns.

src/watcher.js
```diff
@@ -101,11 +101,15 @@ async function getLastBlockToProcess() {
 
 async function main({ sendToQueue }) {
   try {
+    logger.debug('Getting last block')
     const lastBlockToProcess = await getLastBlockToProcess()
```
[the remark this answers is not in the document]
fvictorio: Fixed in 1015ab061646aeec35dbfce5859c96f323404d7b. I just added a bunch of logs in the web3 module, seems more reasonable to have them there.

src/services/logger.js
```diff
@@ -7,6 +7,7 @@ const config =
 const logger = pino({
   enabled: process.env.NODE_ENV !== 'test',
   name: config.name,
+  level: 'debug',
```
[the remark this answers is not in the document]
fvictorio: Thanks for the suggestion. Fixed in c09b7ebe6f1466735dd413881d596931cb3b9705.

src/watcher.js
```diff
@@ -124,10 +128,13 @@ async function main({ sendToQueue }) {
       }
     }
 
+    logger.debug('Updating last processed block')
```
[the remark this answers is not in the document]
fvictorio: Fixed in 30a4d100738d8da8a1f1b5e14c015ffb330998d3.

src/sender.js
```diff
@@ -76,17 +76,22 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
     const gasPrice = await GasPrice.getPrice()
 
     const ttl = REDIS_LOCK_TTL * txArray.length
+
+    logger.debug('Acquiring lock')
     const lock = await redlock.lock(nonceLock, ttl)
 
+    logger.debug('Reading nonce')
     let nonce = await readNonce()
```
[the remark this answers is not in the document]
fvictorio: Fixed in 3402de1073e3c02aaddca769055a5a23269bd09c.

src/sender.js
fvictorio: Releasing the lock is the normal flow of execution, I don't see why it should be a warning or an error.

src/sender.js
fvictorio: Agree, done in 7ca7e58e372fea95bbf56eda1e5bcca4e56f442f.

src/sender.js
fvictorio: This level of detail could be logged at the `trace` level, we could use that instead of having a special flag.

src/sender.js
fvictorio: Fixed in 453cc4bd1f01c9f2f7eccd2bd7bb84963de03444.

src/events/processTransfers/index.js
fvictorio: Done in fbfcd76a613071a26c1d722bac226ffeda8853a2.

src/events/processTransfers/index.js
fvictorio: Yes, you are right, thanks! Fixed in 10870a4a44ad6211f7b6c3a5c22392f9855a9416.

src/events/processTransfers/index.js
fvictorio: Added more logs in b2b75daa0c9acc27a43145461e64963ebad290ae.

src/events/processSignatureRequests/index.js
fvictorio: The `createMessage` function is synchronous. We could add a log here (probably at `trace` level) but it's not something that should get stuck.

src/watcher.js
```diff
@@ -101,11 +101,15 @@ async function getLastBlockToProcess() {
...
+    logger.debug('Getting last block')
     const lastBlockToProcess = await getLastBlockToProcess()
+
     if (lastBlockToProcess <= lastProcessedBlock) {
-      logger.info('All blocks already processed')
+      logger.debug('All blocks already processed')
       return
     }
+
+    logger.debug(`Getting events between block ${lastProcessedBlock + 1} and ${lastBlockToProcess}`)
```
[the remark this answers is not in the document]
patitonar: Fixed on https://github.com/poanetwork/token-bridge/pull/85/commits/962f19ae9fc834eea9e38641510b5b876c092df5

src/watcher.js
```diff
@@ -101,11 +101,15 @@ async function getLastBlockToProcess() {
 
 async function main({ sendToQueue }) {
   try {
+    logger.debug('Getting last block')
     const lastBlockToProcess = await getLastBlockToProcess()
+
     if (lastBlockToProcess <= lastProcessedBlock) {
```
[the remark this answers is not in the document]
patitonar: Done on https://github.com/poanetwork/token-bridge/pull/85/commits/39d6f3ff70faedf787ebc4101b9421def8db3704

src/sender.js
```diff
@@ -76,17 +76,22 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
...
     let insufficientFunds = false
     let minimumBalance = null
     const failedTx = []
 
+    logger.debug(`Sending ${txArray.length} transactions`)
     await syncForEach(txArray, async job => {
       const gasLimit = addExtraGas(job.gasEstimate, EXTRA_GAS_PERCENTAGE)
 
       try {
+        logger.debug(`Sendint transaction with nonce ${nonce}`)
```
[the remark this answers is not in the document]
patitonar: Fixed on https://github.com/poanetwork/token-bridge/pull/85/commits/19ec9327fcfd9cf1b889d309e276878d50b7f423

patitonar:
@akolotov I think all comments were addressed. Unless we are missing something I think we can merge this PR

src/sender.js
akolotov: I thought that `debug` is the lowest level of logging.

akolotov (approved):
Let's do a stress testing for this set of changes and if everything is OK, merge this PR.

akolotov:
Do you have the previous statistic? Can we store it somewhere for reference to find possible degradation?

akolotov:
> One thing that is worth mentioning is that I had to set `MAX_PROCESSING_TIME=25000` to be able to complete processing the 1000 transactions.

That's strange. This is not how I would expect it should work.

fvictorio:
> That's strange. This is not how I would expect it should work.

The watcher processes all the transactions at the same time (even if internally it handles them in chunks of 50), and the watchdog is at the top level of the watcher. I think that's why that happens.

We could move the watchdog at a lower level, or we could make the max processing time a multiple of the number of events that are going to be processed (this also would mean moving it a little lower, but not that much). What do you think?

akolotov:
> We could move the watchdog at a lower level, or we could make the max processing time a multiple of the number of events that are going to be processed (this also would mean moving it a little lower, but not that much). What do you think?

I don't think that it will help if the process hangs in the moment of reading events. 
I think that we need to implement a way to kick the watchdog timer after every chunk of transactions. By doing this the watcher/sender will notify the watchdog that it is still alive.

fvictorio:
That's a great idea, let's do that.

akolotov:
>  let's do that.

My suggestion is to do this under a separate issue.
Currently it is required to confirm that there is no much degradation from the previous version of the token bridge. As soon as we confirm this, the PR could be merged.

fvictorio:
Issue created: https://github.com/poanetwork/token-bridge/issues/110

patitonar:
@akolotov Here are the results of stress testing on `develop` branch:

**Signature Requests**

Tests |1|2|3|AVG
---|---|---|---|---|
count|1000|1000|1000|1000
mean|15431|14850|13376|14552
median|15396|15185|13462|14681
min|5624|5663|5720|5669
max|24536|22354|20412|22434

**Collected Signatures**

Tests |1|2|3|AVG
---|---|---|---|---|
count|1000|1000|1000|1000
mean|9310|9462|9745|9505
median|9755|9476|9961|9730
min|7344|8482|8398|8074
max|10232|10182|10660|10358

If this statistics compared with the [new values](https://github.com/poanetwork/token-bridge/pull/85#issuecomment-433174419) looks good to you, I can add the new ones on `docs/stress-testing.md` so we can use them for future reference.

akolotov:
Thanks @patitonar. I see that there is no significant performance difference between these code sets. So, yes, we can merge these changes.

PR #88, opened 2018-09-27
Fetch gas price after setting the interval

### LINKED ISSUE #86: Fetch gas price immediately
In `services/gasPrice.js`, an interval is set for fetching the gas price. But `setInterval` doesn't run immediately, so the default gas price is used until `updateInterval` milliseconds have passed.

fvictorio:
Closes #86.


akolotov:
@fvictorio do you have ideas why tests were failed?

fvictorio:
No, no at all. I will try re-running it.

fvictorio:
I don't know why this is happening. I will push a revert commit to see if it passes. Please don't merge it.

fvictorio:
I guess this is a problem with Travis: it's failing even with the revert. I also triggered a build in the develop branch and it also failed.

I will try again later. If it persists, I'll try to report the problem.

Issue #87, opened 2018-09-26
Start block parameter skipped when a watcher is started first time


akolotov:
Even if the parameters `HOME_START_BLOCK` and `FOREIGN_START_BLOCK` are configured in `.env` file they will not be used when the bridge is run first time since they are strings where as the code expects them to be integers:

https://github.com/poanetwork/bridge-nodejs/blob/8359556d9e9976a9388720d97de70e1bd7b35af8/src/watcher.js#L27

https://github.com/poanetwork/bridge-nodejs/blob/8359556d9e9976a9388720d97de70e1bd7b35af8/src/watcher.js#L106-L112

Scenario to reproduce:
1. Deploy the bridge contracts in the `erc-to-erc` mode. Store `bridgeDeploymentResults.json` created after the deployment.
2. Wait for few blocks produced
3. Transfer tokens from `Foreign` network
4. Wait for few blocks produced
5. Configure `.env` of the bridge instance and points `HOME_START_BLOCK` and `FOREIGN_START_BLOCK` with values inserted in `bridgeDeploymentResults.json` after the contracts deployment.
6. Run the bridge processes with empty Redis DB.
As soon as the `watcher-erc-affirmation-request` process started following appeared in the terminal:
```
"Found 0 Transfer events"
```
and the corresponding requests to relay assets (`transfer` calls) does not handled by the bridge.

akolotov:
Addressed in #103

PR #89, opened 2018-09-30
Integration with RSK

akolotov:
As per adaptation to use the POA bridge with RSK network the following changes were made:
1. gas price must be specified in Wei rather than GWei
2. added CLI to access quickly to JSON-RPC calls like `getTransactionByHash`, `getReceipt`, `getBlockByNumber` etc.
3. Fix for #87 
4. Script to generate transaction modified to use gas price from environment variables
5. extra logs added to help with issue trubleshooting
6. FORCE_NONCE_UPDATE introduced to force the sender to get the NOnce from the RPC node always.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 node_modules
 .env
+.env.*
 .idea
 .nyc_output
 coverage/
 logs/
+*.log
--- a/scripts/getNonce.sh
+++ b/scripts/getNonce.sh
@@ -1,0 +1,33 @@
+#9!/usr/bin/env bash
+
+if [ -f .env ]; then
+  source .env
+fi
+
+if [ "$1" == "" ]; then
+  echo "Account is not specified" 1>&2
+  exit 1
+fi
+
+account=$1
+
+remoteUrl=""
+
+if [ "$2" == "" ]; then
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -1,6 +1,8 @@
 node_modules
 .env
+.env.*
```
fvictorio: If we do this, maybe we should also add a `!.env.example` below. It's not necessary, since that file is already in the git index, but it would make explicit that it shouldn't be ignored.

scripts/getNonce.sh
```diff
@@ -0,0 +1,33 @@
+#9!/usr/bin/env bash
```
fvictorio: That `9` is a typo, isn't it?

scripts/sendUserTxToErcForeign.js
```diff
@@ -67,6 +68,7 @@ async function main() {
       const gasLimit = await poa20.methods
```
patitonar: In order to be re-assigned, `gasLimit` should be declared with `let` instead of `const`

patitonar:
CI failed because of the linter. You can fix the code format manually or run `eslint . --fix`

akolotov:
Not applicable

PR #91, opened 2018-10-01
Set persistent storage for rabbitmq and redis services

ArseniiPetrovich:


docker-compose.yml
```diff
@@ -2,12 +2,25 @@ version: "3"
 services:
   redis:
     image: "redis"
```
akolotov: Should we specify exact version of `redis` here? Similarly as it was done in https://github.com/poanetwork/bridge-nodejs/pull/88/files#diff-ac440e3f7ab9ccaac9aeab32ac91b8b3?

docker-compose.yml
```diff
@@ -2,12 +2,25 @@ version: "3"
...
+    command: ["redis-server", "--appendonly", "yes"]
+    hostname: redis
+    restart: on-failure
+    volumes:
+      - ~/bridge_data/redis:/data
```
akolotov: What component is responsible for creation `~/bridge_data` on the host system?

docker-compose.yml
ArseniiPetrovich: Yes, it is definitely a good idea to explicitly specify a version of `redis` image.

docker-compose.yml
ArseniiPetrovich: @akolotov Docker creates all the folder tree automatically while building images with attached drives. Of course, user that build images should have sufficient permissions.

PR #95, opened 2018-10-04
Add default gas price update interval

### LINKED ISSUE #83: Incorrect behavior if *_GAS_PRICE_UPDATE_INTERVAL is not defined
If there is no gas price oracle for particular network it seems logical to not define any of these parameters:
```
*_GAS_PRICE_ORACLE_URL=
*_GAS_PRICE_SPEED_TYPE=
*_GAS_PRICE_UPDATE_INTERVAL=
```
At the same time it makes sense to configure `*_GAS_PRICE_FALLBACK` assuming that it will be used if the gas price values will not be available from the contract.

If such configuration is used the following code does not work properly since `updateInterval` is `undefined`.

```
  fetchGasPriceInterval = setInterval(async () => {                                                                                                                                                         
    const gasPrice = await fetchGasPrice({                                                                                                                                                                  
      bridgeContract,                                                                                                                                                                                       
      oracleFn: () => fetchGasPriceFromOracle(oracleUrl, speedType)                                                                                                                                         
    })                                                                                                                                                                                                      
    cachedGasPrice = gasPrice || cachedGasPrice                                                                                                                                                             
    logger.info(`Updated gas price: ${cachedGasPrice}`)                                                                                                                                                     
  }, updateInterval)                                                                                                                                                                                        
```

Consider to initialize `updateInterval` in some default value if `*_GAS_PRICE_UPDATE_INTERVAL` is not configured.

patitonar:
Closes #83 

Adds a default value for gas price update interval to be used if the value is not defined on `.env`. I used `600000` which is the value defined on `.env.example`, let me know if you think of a better value.

### DIFF
--- a/e2e/parity/Dockerfile
+++ b/e2e/parity/Dockerfile
@@ -1,4 +1,4 @@
-FROM parity/parity:stable
+FROM parity/parity:v1.11.11
 
 WORKDIR /stuff
 
--- a/src/services/gasPrice.js
+++ b/src/services/gasPrice.js
@@ -65,7 +65,7 @@ async function start(chainId) {
   let bridgeContract = null
   let oracleUrl = null
   let speedType = null
-  let updateInterval = null
+  let updateInterval = 600000
   if (chainId === 'home') {
     bridgeContract = homeBridge
     oracleUrl = HOME_GAS_PRICE_ORACLE_URL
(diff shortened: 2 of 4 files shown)


akolotov:
Sorry for confusion. I meant that you need to add a test where this parameter is not initialized in `.env` file. This test should demonstrate that the sender works as expected - gets the value after the default timeout. One test to get the value from the oracle, another test to get the value from the contract.

patitonar:
@akolotov Fixed the code, and added tests for this. The tests checks that the interval is created with the correct timeout if `.env` variable if defined or not. Calling method from contract if oracle fails was already covered on tests.

akolotov:
@patitonar Do you know what is the reason for failing the Travis CI build?

akolotov:
> We should review the configuration of the docker image

Let's focus on the Bridge UI first. We need to prepare it for a new DAI chain.

patitonar:
@akolotov CI is fixed now

PR #97, opened 2018-10-08
fix typos

andogro:


andogro:
This will be included with #96

PR #102, opened 2018-10-12
Fix issue with start block parameters

### LINKED ISSUE #87: Start block parameter skipped when a watcher is started first time

fvictorio:
Closes #87.

Note that the names `HOME_START_BLOCK` and `FOREIGN_START_BLOCK` may be misleading: they are being interpreted as "last processed", so it will actually start from `HOME_START_BLOCK + 1`.

@akolotov do you think we should fix this? We can change the variable names to better reflect what they do, or we can fix it by subtracting 1 to them when they are used as last processed block numbers.

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -1,6 +1,6 @@
 require('dotenv').config()
 const path = require('path')
-const { toBN } = require('web3').utils
+const { BN, toBN } = require('web3').utils
 const { connectWatcherToQueue, connection } = require('./services/amqpClient')
 const { getBlockNumber } = require('./tx/web3')
 const { redis } = require('./services/redisClient')
@@ -21,11 +21,14 @@ const processCollectedSignatures = require('./events/processCollectedSignatures'
 const processAffirmationRequests = require('./events/processAffirmationRequests')(config)
 const processTransfers = require('./events/processTransfers')(config)
 
+const ZERO = toBN(0)
+const ONE = toBN(1)
+
 const web3Instance = config.web3
 const bridgeContract = new web3Instance.eth.Contract(config.bridgeAbi, config.bridgeContractAddress)
 const eventContract = new web3Instance.eth.Contract(config.eventAbi, config.eventContractAddress)
 const lastBlockRedisKey = `${config.id}:lastProcessedBlock`
-let lastProcessedBlock = config.startBlock || 0
+let lastProcessedBlock = BN.max(config.startBlock.sub(ONE), ZERO)
 
 async function initialize() {
   try {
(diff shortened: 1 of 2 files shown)


fvictorio:
Done. To simplify things, I made sure that `lastProcessedBlock` is always a BN. Otherwise we are always guessing if it is a string, a number or a bignumber.

akolotov:
You need to merge to `develop` (please create it) not to `master`.

src/watcher.js
```diff
@@ -60,12 +64,12 @@ async function runMain({ sendToQueue }) {
...
 async function getLastProcessedBlock() {
   const result = await redis.get(lastBlockRedisKey)
-  lastProcessedBlock = result ? Number(result) : lastProcessedBlock
+  lastProcessedBlock = result ? toBN(result) : lastProcessedBlock
 }
 
 function updateLastProcessedBlock(lastBlockNumber) {
   lastProcessedBlock = lastBlockNumber
-  return redis.set(lastBlockRedisKey, lastProcessedBlock)
+  return redis.set(lastBlockRedisKey, lastProcessedBlock.toString())
```
akolotov: if I call `reset-lastBlock.sh` in CLI which type will be used to write the value to the database?

fvictorio: `reset-lastBlock.sh` updates the last block processed in redis. So callingit with `11` is the same as setting `START_BLOCK` to 10 and clearing the DB. Does that make sense?

fvictorio:
> You need to merge to `develop` (please create it) not to `master`.

Oh, sorry. I just opened #103.

src/watcher.js
akolotov: I meant which type of the value will be stored there: number of string?

fvictorio: Oh, sorry. All values in redis are strings as far as I know. When we obtain the value from redis, we convert it to BN.

akolotov: ok. thanks

PR #106, opened 2018-10-18
Add watchdog for watchers

### LINKED ISSUE #105: Run a watchdog timer in watchers
The watchers of a deployed bridge instance have had issues in which they stop listening. We are not sure about what's happening, but to alleviate the issue we need to add a watchdog that kills the process if the watcher function doesn't return after a given amount of time.

Killing the process has an edge case where it causes a problem: if it's killed after the jobs are sent to rabbit but before the last processed block is updated in redis, the transactions involved may get sent twice. This is very unlikely, and the only downside is making the validators waste gas, so it's acceptable for now.

fvictorio:
Closes #105.

The environment variable used for the max time is `MAX_PROCESSING_TIME`. If it's not set, the watchdog is not used.

I also moved the exit codes to the constants files, so that we know what each of them means.


akolotov:
Please change the target branch to `develop`

.env.example
```diff
@@ -28,6 +28,8 @@ REDIS_LOCK_TTL=1000
 HOME_START_BLOCK=
 FOREIGN_START_BLOCK=
 
+MAX_PROCESSING_TIME=600000
```
akolotov: Is my understanding correct and 10 min is too much for the watchdog timer? Could you estimate max time expected to handle one iteration of watcher?

.env.example
```diff
@@ -28,6 +28,8 @@ REDIS_LOCK_TTL=1000
...
+
```
akolotov: Please update README file as well with the new parameter.

akolotov:
I believe we need to expanse these changes on the senders as well. What do you think?

.env.example
fvictorio: I guess it depends on the number of events processed at the same time. I will try sending several transactions to the kovan-sokol bridge and see how much time is spent processing them.

fvictorio: Sent 10 transactions to Sokol. 2 of them were processed in one block, 8 in another. The time difference between the first log, and the log of the sender receiving the last transactions to send was 3 seconds (that is Sokol's block time, isn't it?).

The time is going to depend on the chain, but I guess one minute should be enough in the vast majority of cases (and it can be configured, anyway).

akolotov: Hmm. I think the timout should be comparable with the polling time. I think 3 or 4 times more than the polling time of the corresponding network. What do you think?

fvictorio: Yes, that sounds reasonable. You mean using that as the example, or that the timeout should be computed from the polling time?

akolotov: Yes, let's set the timeout in the example to be 4 times more that the longest polling time and provide these recommendations in README.

fvictorio: Please check 3171c1e6b80ece4c289d677d4081c1da213caa27.

src/sender.js
```diff
@@ -36,7 +38,16 @@ async function initialize() {
...
+            logger.fatal('Max processing time reached')
+            process.exit(EXIT_CODES.MAX_TIME_REACHED)
+          })
+        }
+
+        return main(options)
+      }
     })
   } catch (e) {
     logger.error(e.message)
```
akolotov: Should `process.exit(1)` also use a constant from `EXIT_CODES`?

src/watcher.js
```diff
@@ -51,7 +54,14 @@ async function initialize() {
 async function runMain({ sendToQueue }) {
```
akolotov: Should `process.exit(1)` in the line 50 use a constant from `EXIT_CODES` as well?

src/watcher.js
```diff
@@ -7,7 +7,8 @@ const { redis } = require('./services/redisClient')
 const logger = require('./services/logger')
 const rpcUrlsManager = require('./services/getRpcUrlsManager')
 const { getRequiredBlockConfirmations, getEvents } = require('./tx/web3')
-const { checkHTTPS } = require('./utils/utils')
+const { checkHTTPS, watchdog } = require('./utils/utils')
+const { EXIT_CODES } = require('./utils/constants')
 
 if (process.argv.length < 3) {
   logger.error('Please check the number of arguments, config file was not provided')
```
akolotov: Should `process.exit(1)` use a constant from `EXIT_CODES` as well?

src/sender.js
```diff
@@ -10,11 +10,12 @@ const { getNonce, getChainId } = require('./tx/web3')
...
+  privateKeyToAddress,
   syncForEach,
   waitForFunds,
-  privateKeyToAddress
+  watchdog
 } = require('./utils/utils')
-const { EXTRA_GAS_PERCENTAGE } = require('./utils/constants')
+const { EXIT_CODES, EXTRA_GAS_PERCENTAGE } = require('./utils/constants')
 
 const { VALIDATOR_ADDRESS_PRIVATE_KEY, REDIS_LOCK_TTL } = process.env
```
akolotov: Should `process.exit(1)` in the line 26 use a constant from `EXIT_CODES` as well?

src/sender.js
```diff
@@ -32,6 +33,8 @@ const nonceLock = `lock:${config.id}:nonce`
 const nonceKey = `${config.id}:nonce`
 let chainId = 0
 
+const maxProcessingTime = process.env.MAX_PROCESSING_TIME
```
akolotov: We need to define a default value which will be used if `MAX_PROCESSING_TIME` was missed by some reason in the configuration file.

src/watcher.js
```diff
@@ -30,6 +31,8 @@ const eventContract = new web3Instance.eth.Contract(config.eventAbi, config.even
 const lastBlockRedisKey = `${config.id}:lastProcessedBlock`
 let lastProcessedBlock = BN.max(config.startBlock.sub(ONE), ZERO)
 
+const maxProcessingTime = process.env.MAX_PROCESSING_TIME
```
akolotov: We need to define a default value which will be used if `MAX_PROCESSING_TIME` was missed by some reason in the configuration file.

src/sender.js
akolotov: What risks can you see if we enable the watchdog by default? In this case the user could disable the watchdog by zeroing the configuration parameter explicitly.

fvictorio: That is a bit complicated. Javascript treats 0 as a falsy value, so doing `0 || 10` evaluates to `10`. 

On the other hand, the environment variables are strings, so `'0' || 10` does evaluate to `'0'`, but it's not good to rely on that; we could easily "break" it if we add environment variables validation and casting, like we did in `poa-bridge-contracts/deploy`.

A possible solution is to use a negative value to disable it. That will work both with strings and with numbers: `'-1' < 0` is equivalent to `-1 < 0`.

src/sender.js
fvictorio: Done in 3d706ddaf7e108b1ce2fe4756364236d845b7a24.

src/sender.js
akolotov: I do not understand this.
Could you use something like:
```
if (process.env.MAX_PROCESSING_TIME === '0') {
...
} else {
...
}
```

And later, when you _add environment variables validation and casting_ you will change this.

fvictorio: That can work. This is what I did (see f38688c869be9fcd5e8dceb36b31637769a6f802):
1. If value is explicitly set to 0, the watchdog is not used.
2. If the value is not set, it is computed using 4 * maxPollingTime
3. Otherwise, the set value is used

I didn't want to add this logic twice (in the watcher and the sender), so I moved it to the config. And since the senders weren't including the base config, I also added it to them. 

What do you think?

akolotov:
Merge the branch `support-erc20-native-#81` in order to fix the issue with TravisCI and as soon as Travis does not report about any issue, merge this PR.

PR #113, opened 2018-11-02
Changed docker-compose file.

VladimirNovgorodov:


akolotov:
@akolotov @fvictorio please take a look at this PR

pertsev:
@VladimirNovgorodov Please, add CPU and RAM limits for containers.

.dockerignore
```diff
@@ -1,3 +1,11 @@
-node_modules
```
patitonar: I think `node_modules` should be ignored. I see there was a similar conversation on this comment https://github.com/poanetwork/token-bridge/pull/109#discussion_r227427251

akolotov:
@VladimirNovgorodov could you share a forecast when the comments raised in this PR are addressed?
We need to provide responses on the issues found during the security audit on this week. Is it possible?

@phahulin, fyi

VladimirNovgorodov:
Yes.  I'm working on it.

VladimirNovgorodov:
> @VladimirNovgorodov Please, add CPU and RAM limits for containers.
Docker-compose version 3 that we use only supports these limits in swarm mode. https://forums.docker.com/t/memory-limit-in-version-3-compose-files-without-using-swarm/37864. 
In this case, I need to know if we can use Docker-compose file version 2 or maybe some other options.

patitonar:
@VladimirNovgorodov should we close this PR in favor of #116 ?

phahulin:
Yes

PR #116, opened 2018-11-13
Security review fixes

ArseniiPetrovich:
This one is similar to #113, but also takes into account all review remarks.


docker-compose.yml
```diff
@@ -1,26 +1,134 @@
...
+    restart: on-failure
+  bridge_request:
+    cpus: 0.1
+    mem_limit: 500m
+    build: .
+    command: 'true'
+    env_file: ./.env
+    environment: 
+      - NODE_ENV=production 
+      - VALIDATOR_ADDRESS=${VALIDATOR_ADDRESS:-0x8fd379246834eac74B8419FfdA202CF8051F7A03}
```
phahulin: I think it's better not to use any default values here.
Applies to all occurences of `VALIDATOR_ADDRESS` and `VALIDATOR_ADDRESS_PRIVATE_KEY` below

patitonar:
Should this PR target `develop` instead of `master` branch @akolotov ?

docker-compose.yml
```diff
@@ -1,26 +1,134 @@
...
+    restart: on-failure
+  bridge_request:
+    cpus: 0.1
+    mem_limit: 500m
+    build: .
+    command: 'true'
+    env_file: ./.env
+    environment: 
+      - NODE_ENV=production 
+      - VALIDATOR_ADDRESS=${VALIDATOR_ADDRESS}
```
akolotov: I think we don't need `VALIDATOR_ADDRESS` here since the address will be generated automatically based on the private key provided in the next line.
@patitonar please confirm.

docker-compose.yml
```diff
@@ -1,26 +1,134 @@
...
+      - net_rabbit_bridge_affirmation
+  bridge_senderhome:
+    cpus: 0.1
+    mem_limit: 500m
+    build: .
+    command: 'true'
+    env_file: ./.env
+    environment: 
+      - NODE_ENV=production 
+      - VALIDATOR_ADDRESS=${VALIDATOR_ADDRESS}
```
akolotov: I think we don't need `VALIDATOR_ADDRESS` here since the address will be generated automatically based on the private key provided in the next line.

docker-compose.yml
```diff
@@ -1,26 +1,134 @@
...
-    environment:
-      - NODE_ENV=production
-      - QUEUE_URL=amqp://rabbit
-      - REDIS_URL=redis://redis
-    command: "true"
-    restart: on-failure
\ No newline at end of file
+    environment: 
+      - NODE_ENV=production 
+      - VALIDATOR_ADDRESS=${VALIDATOR_ADDRESS}
```
akolotov: I think we don't need `VALIDATOR_ADDRESS` here since the address will be generated automatically based on the private key provided in the next line.

akolotov:
> Should this PR target develop instead of master branch @akolotov ?

Yes, it is correct. We need to use `develop` branch. As soon as the PR is merged @patitonar, I think, we need to run stress testing to see if the CPU and memory limits declared by the docker-compose configuration file are enough.

docker-compose.yml
patitonar: That's correct. `VALIDATOR_ADDRESS` is generated automatically using the private key.

ArseniiPetrovich: Hm. I might be mistaking, but it won't work if not specifying the `VALIDATOR_ADDRESS`. However, let me check it first.

ArseniiPetrovich:
Should be targeted to different branch.

akolotov:
@ArseniiPetrovich are you going to open another PR for the `develop` branch?

ArseniiPetrovich:
> @ArseniiPetrovich are you going to open another PR for the `develop` branch?

Yes, @akolotov. Check #119, please

PR #133, opened 2019-02-21
Add boundaries to oracle's gas price

### LINKED ISSUE #132: Attack to the gas price oracle could lead to drying out of a validator balance
Currently the token bridge instances uses the gas price oracle to get the gas price for transactions directed to the ETH Mainnet. If the oracle will be hacked or the connection to the oracle will be spoofed, the bridge instance could get any value in the response.

https://github.com/poanetwork/token-bridge/blob/7ccf71df5445f10c1f5f3cc23fe9cf5095939845/src/services/gasPrice.js#L34-L42

Since there is no check if the suggested gas price is within some limits, the attacker could use a huge value to dry out the validator's account balance or vice versa - a very small value to block the transactions from fast mining.

Consider to introduce a check after the following line:
https://github.com/poanetwork/token-bridge/blob/7ccf71df5445f10c1f5f3cc23fe9cf5095939845/src/services/gasPrice.js#L37
and fail if the response is not within limits.

patitonar:
Closes #132

### DIFF
--- a/src/services/gasPrice.js
+++ b/src/services/gasPrice.js
@@ -7,7 +7,7 @@ const logger = require('../services/logger').child({
   module: 'gasPrice'
 })
 const { setIntervalAndRun } = require('../utils/utils')
-const { DEFAULT_UPDATE_INTERVAL } = require('../utils/constants')
+const { DEFAULT_UPDATE_INTERVAL, GAS_PRICE_BOUNDARIES } = require('../utils/constants')
 
 const HomeABI = bridgeConfig.homeBridgeAbi
 const ForeignABI = bridgeConfig.foreignBridgeAbi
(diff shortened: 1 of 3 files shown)


src/services/gasPrice.js
```diff
@@ -31,12 +31,20 @@ const foreignBridge = new web3Foreign.eth.Contract(ForeignABI, FOREIGN_BRIDGE_AD
...
+
 async function fetchGasPriceFromOracle(oracleUrl, speedType) {
   const response = await fetch(oracleUrl)
   const json = await response.json()
   const gasPrice = json[speedType]
   if (!gasPrice) {
     throw new Error(`Response from Oracle didn't include gas price for ${speedType} type.`)
+  } else if (!gasPriceWithinLimits(gasPrice)) {
+    throw new Error(
+      `Response from Oracle included a gas price out of boundaries ${gasPrice} for ${speedType} type.`
```
akolotov: And, please correct me if I am wrong, we will use cached version of the gas price? Why do you think it is better than use the limit instead? For example, if the gas price oracle returns 0.5 `gwei`, we will use 1 `gwei`, if the oracle returns 300 `gwei` we will use 250 `gwei`. The reason is that if we have a real oracle's response (rather than compromised), these values will not impact much the transactions. Especially in case if the gas price goes up suddenly - the usage of the limit value will still allow to transactions included to the block whereas the fallback value will slow the transactions down.

patitonar: I think you are right, we should use the limit

patitonar: Done  https://github.com/poanetwork/token-bridge/pull/133/commits/7654da6df98826ec28f0dbca74e8c13e220cc9b4

akolotov: thanks!

PR #136, opened 2019-03-22
Use chainId on sender instead of networkId

### LINKED ISSUE #134: NetworkId is confused with chainId
```getChainID``` in ```\src\tx\web3.js``` returns network id instead of chainId.

This can lead to ```invalid sender``` errors for certain network config.

 I fixed it on my side by wrapping rpc ```eth_chainId```

patitonar:
Closes #134

### DIFF
--- a/e2e/package-lock.json
+++ b/e2e/package-lock.json
@@ -14,14 +14,14 @@
       }
     },
     "ajv": {
-      "version": "5.5.2",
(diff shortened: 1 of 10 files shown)


patitonar:
I had to update the parity version we were using on the e2e tests. `eth_chainId` method was implemented on version `Parity-Ethereum v2.2.1 (2018-11-15)`

Issue #78, opened 2018-09-06
Check that the validator is valid when the worker starts


fvictorio:
See [these comments](https://github.com/poanetwork/bridge-nodejs/pull/71#discussion_r215230667). This makes sense for all workers, so this should be part of some common initialization (maybe `start-worker.sh` could make a call to `check-validator.js` before starting?)

PR #145, opened 2019-05-03
Add support for UserRequestForAffirmation event on erc-to-erc mode

### LINKED ISSUE #141: Watcher for UserRequestForAffirmation events in the erc20-to-erc20 bridge mode
Current implementation of the `erc20-to-erc20` bridge mode assumes that pure ERC20 token is used on the Foreign side. So, when the tokens are relayed to the Home side, the method `transfer` is invoked by the user and the relay request is handled by the TokenBridge oracle without involvement the bridge contract on the Foreign side. It is inconvenient due to following:
- there is no way to revert the request if it exceeds the limits.
- there is no way to halt the bridge operations

Historically, such kind of functionality was suggested for already deployed ERC20 tokens to enhance them through the bridging ability and described shortcomings were considered as insignificant in comparison with the value such solution could bring. But for there are the cases for the bridge when it is going to be used with just deployed tokens. In this case the token could be extended by ERC677 standard in order to address issues described above. 

That's why the PR https://github.com/poanetwork/poa-bridge-contracts/issues/175 was created in the bridge contracts repo. In order to allow the token bridge oracle handle such relay request the watcher to catch the `UserRequestForAffirmation` events must be run if the bridge is in `erc20-to-erc20` mode. 

As per the initialization procedure the token bridge oracle must check whether the ERC20 token contract has the `bridgeContract` method and it points to the same birdge contract the oracle must monitor. If so, the watcher will look for the `UserRequestForAffirmation` events. If not, the `Transfer` event will being caught.

patitonar:
Closes #141 

Since the call to check the `bridgeContract` method couldn't be made on the config files because of their synchronous nature,  I added a new script that is going to be called on `start-worker.sh` before the watcher start and will pass an object with information to the watcher.

The idea of this script is to perform any kind of check it is required for the watcher to work correctly.
In this case it will check which event should listen to on foreign side on `erc-to-erc`. 

In the future this script could be extended to check that the validator is valid https://github.com/poanetwork/token-bridge/issues/78, or to get the bridge mode (so there is no need to provide it from env variable), get the erc20 token address from foreign (so we can avoid this env variable too and only need the home and foreign bridge) and even to get the start block from which the validator was added and avoid those env variable too.

Also I had to update parity version used on e2e scripts because previous one was not longer available.

I think we should add an e2e test for this bridge mode too, but we can create an issue for this and address it in a separate PR.

### DIFF
--- a/config/affirmation-request-watcher.config.js
+++ b/config/affirmation-request-watcher.config.js
@@ -1,11 +1,25 @@
 require('dotenv').config()
 const baseConfig = require('./base.config')
 const erc20Abi = require('../abis/ERC20.abi')
+const { ERC_TYPES } = require('../src/utils/constants')
+
+const initialChecksJson = process.argv[3]
+let initialChecks
+try {
+  initialChecks = JSON.parse(initialChecksJson)
+} catch (e) {
+  throw new Error('Error on decoding values from initial checks.')
+}
+
+if (baseConfig.id === 'erc-erc' && initialChecks.foreignERC === ERC_TYPES.ERC677) {
+  baseConfig.id = 'erc677-erc677'
+}
 
 const id = `${baseConfig.id}-affirmation-request`
 
 module.exports =
-  baseConfig.id === 'erc-erc' || baseConfig.id === 'erc-native'
+  (baseConfig.id === 'erc-erc' && initialChecks.foreignERC === ERC_TYPES.ERC20) ||
+  baseConfig.id === 'erc-native'
     ? {
         ...baseConfig.bridgeConfig,
         ...baseConfig.foreignConfig,
--- a/config/affirmation-request-watcher.config.js
+++ b/config/affirmation-request-watcher.config.js
@@ -4,6 +4,11 @@ const erc20Abi = require('../abis/ERC20.abi')
 const { ERC_TYPES } = require('../src/utils/constants')
 
 const initialChecksJson = process.argv[3]
+
+if (!initialChecksJson) {
+  throw new Error('initial check parameter was not provided.')
+}
+
 let initialChecks
 try {
   initialChecks = JSON.parse(initialChecksJson)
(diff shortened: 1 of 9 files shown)


config/affirmation-request-watcher.config.js
```diff
@@ -1,11 +1,25 @@
 require('dotenv').config()
 const baseConfig = require('./base.config')
 const erc20Abi = require('../abis/ERC20.abi')
+const { ERC_TYPES } = require('../src/utils/constants')
+
+const initialChecksJson = process.argv[3]
```
akolotov: could we check if `argv[3]` exists in order to log proper message if `affirmation-request-watcher` was run without this parameter?

scripts/initialChecks.js
```diff
@@ -0,0 +1,33 @@
...
+})
+const Web3 = require('web3')
+const ERC677BridgeTokenABI = require('../abis/ERC677BridgeToken.abi')
+const { ERC_TYPES } = require('../src/utils/constants')
+
+async function initialChecks() {
+  const { ERC20_TOKEN_ADDRESS, BRIDGE_MODE, FOREIGN_RPC_URL, FOREIGN_BRIDGE_ADDRESS } = process.env
+  const result = {}
+
+  if (BRIDGE_MODE === 'ERC_TO_ERC' && process.argv[2].includes('affirmation-request-watcher')) {
```
akolotov: What benefit to specify that the json structure must be generated for `affirmation-request-watcher` only is? It seems that this script too tied to the rest of functionality, so if the watcher name is changed, the changes in this file must be done as well. What do you think?

config/affirmation-request-watcher.config.js
patitonar: Yes I think it makes sense. Added a check here https://github.com/poanetwork/token-bridge/pull/145/commits/ff56ec992942160009fdc74367dc0e3b178a6c59

scripts/initialChecks.js
patitonar: The idea was to avoid other watchers to do those checks since they don't need it. But I agree with you that this is too tied and it won't have a big impact on other watchers. Updated here https://github.com/poanetwork/token-bridge/pull/145/commits/f2610a95161c55007c69f0ff34d3788ac69f22f2

